PR #22, opened 2017-09-25
Add angular skeleton

anderaus:
Angular skeleton app. Intended to be integrated with @einari's backend example. As mentioned to Einar, Angular 2 (now 4, actually) is quite a handful setup-wise - so I've found it very useful in other projects to base everything on the official best practices and use the angular CLI tool (which also means using typescript and sass/scss).

I've added a short readme - should be enough to get the bits up and running. Feel free to have a go, @karolikl . When Einar's backend is more complete, I'll try calling some of his stuff and get something working end-to-end. Feel free to comment on anything!


karolikl:
Looks good!

PR #173, opened 2017-10-01
Add build status for Example and VolunteerReporting

tksoftware:
Link the build status badges against Roxanne appveyor account


tksoftware:
I see that I updated the wrong reedme file I will make a new PR moving the status when I return to my computer

Issue #205, opened 2017-10-03
Deleted empty documentation pages, consider including them if necessary


karolikl:
Deleted from architecture: 
- Event Driven
- Event Processor
- Infrastructure
- Integration
- Kubernetes
- Multitenancy
- On Premise
- Overview
- Partitioning strategy
- Runtime Environment
- Secrets
- SMS Gateway

Deleted from contribution:
- Contributors License Agreement
- Debugging
- Logging
- Visual Studio Code
- Testing
- Pull Requests

Issue #133, opened 2017-09-30
CaseReportReceived


Molokai:
```
        public Guid CaseReportId { get; set; }
        public Guid DataCollectorId { get; set; }
        public Guid HealthRiskId { get; set; }
        public string Origin { get; set; }
        public int NumberOfMalesUnder5 { get; set; }
        public int NumberOfMalesAged5AndOlder { get; set; }
        public int NumberOfFemalesUnder5 { get; set; }
        public int NumberOfFemalesAged5AndOlder { get; set; }
        public double Longitude { get; set; }
        public double Latitude { get; set; }
        public DateTimeOffset Timestamp {get; set;}
```

Existing entry:

### Description
Event emitted by VolunteerReporting when a DataCollector (reporter / volunteer) has submitted a case report that has been successfully parsed (for example submitted a correctly formatted SMS).

The event contains the following information:
- ID of the DataCollector
- Timestamp of the submission
- If it is a single case report or an aggregate report (see #44 #45 #46 )
- Aggregated list of elements per unique composition of health risk, gender and age
```
caseList : [
        {"healthRisk" : "51BBA12C-7E52-4DE8-8D0C-114E32CFBD04",
         "gender" : "male",
         "age" :  2,
         "numberOfCases" : 3},

        {"healthRisk" : "51BBA12C-7E52-4DE8-8D0C-114E32CFBD04",
         "gender" : "female",
         "age" : 1,
         "numberOfCases" : 2,}
         ]

TODO: HOW is age represented for the Aggregrate Case Report where the columns are only less than 5 or greater than 5.

FOR EXAMPLE:
caseList : [
        {"healthRisk" : "51BBA12C-7E52-4DE8-8D0C-114E32CFBD04",
         "gender" : "male",
         "ageBracket" :  "Below 5",
         "numberOfCases" : 3},

        {"healthRisk" : "51BBA12C-7E52-4DE8-8D0C-114E32CFBD04",
         "gender" : "female",
         "ageBracket" : "Above 5",
         "numberOfCases" : 2}
         ]


- Supplemental data?  (could include geo tagging if available. TBD)

### Class

Velcrow81:
Should we have a different event for aggregated events? I.e. CaseReportsReceived? That way, they can contain only the information relevant for the type of case?

Velcrow81:
Can one text message contain multiple single case reports? If so, I think the text parsing must be updated. I see that it says numberOfCases in both multiple and single case report.

[...]

PR #209, opened 2017-10-05
Event clarification and also guidance on not defaulting to the DRY principle

einari:
This should clarify some of the questions raised in #133.

Its also related to #205 - for more details and more to come.

### DIFF
--- a/Documentation/Architecture/events.md
+++ b/Documentation/Architecture/events.md
@@ -132,7 +132,8 @@ This does not mean that you should be keeping full entities on the event - still
 Be aware of this and model accordingly.
 
 Important to keep in mind that the snapshot of a state is a series of events, these together form
-the snapshot. This should reduce what you keep on them but as a whole they help guarantee the
+the snapshot. This should reduce what you keep on them but as a whole they help guarantee the result
+when a replay would occur.
 
 In addition to events being modelled correctly and the processor doing it right, ordering is an important
 aspect of dealing with events. The underlying system guarantees ordering of events when replaying and also
--- a/Documentation/Architecture/events.md
+++ b/Documentation/Architecture/events.md
@@ -6,11 +6,15 @@ author: einari
 ---
 # Events
 
-Unlike representing the user intention through a [command](./commands.md) or similar, an event
-represents the actual state change in the system - the thing that happpened. Its name should be
-past-tense, reflecting that it has happened.
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”,
+“RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in
+[RFC 2119](https://tools.ietf.org/html/rfc2119).
 
-By capturing things in events we get new opportunities in representing the actual truth to what is
+Events represents the actual state change happening in a system. Rather than just saving the state,
+you represent the change with an event. It is not technical in nature - but represents the domain
+language - the [ubiquitous language](https://en.wikipedia.org/wiki/Domain-driven_design) in your domain.
+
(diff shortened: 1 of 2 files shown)


roarfred:
About events and indempotecy... does it mean the same event should be allowed to be raised several times, successivly, or does it (in addition) mean that you should allow the full event store to be replayed from any point to the end?

einari:
You don't have to emit an event more than once. For idempotency it is important that when replaying you are able to reach the same result as you did when it was running. This affects how you model the event, but even more how you implement the processor

karolikl:
I'd like to rephrase Roars question a bit: What should happen IF we receive the same event twice (let's say because of an error in the infrastructore? 

Example: We receive the TextMessageReceived event, which stores the text message, parses it and creates a case report. If we receive the same event again, should we: 
1. store the message again, process it again and create a new case report (leaving us with a duplicate)
or should we
2. identify that this event has already been processed and update the existing case report?

einari:
At runtime you won't get the event twice. The replay is typically used when targeting a new read models repository. For event processors that would generate an email, sms or similar you'd implement a different method. Will announce that when is ready. 

But rule of thumb ; events won't happen  twice - it will be guaranteed once and once only. Ordering will also be guaranteed. 

In other words this is not something to worry about. 

Should I clarify this in the text somehow? It's kinda in there if you read it all. Reorder it? Rearrange? Have a checklist at the top of things you should think about - or a table? Very open for input here :)

roarfred:
What puzzled me was the requirement for idempotency, as I see this could play out in three different scenarios:
1) The same event is received twice, don't let it affect your state (data store)
2) Playback from start at the initial state, make sure you always end up at the same state
3) Playback from an arbitrary point in time, at the current state. Make sure your state is not changed when all events are re-processed.

The last one is most about "what if I receive an event telling me something has changed, and this something doesn't exist in my state/data store?". I assume we should always be able to survive, even if the incoming event doesn't seem to make any sense. In the end, there is no way to bite back at the sender, is there?

einari:
I'll rephrase it completely to avoid confusion. Maybe even remove it.

Idempontency is often something referred to in event sourcing - but the underlying system pretty much takes the problem away. The point of the paragraph is to make one aware of the concept. 

Let this be open till I fix it :)

roarfred:
My comments were never meant to be an obstacle for the pull request, just found a spot to ask questions. Please excuse my ramblings and let the pull request go through, it clearly adds value to the master branch without these questions answered!

einari:
:) no worries - good to get the message clear. Don't want confusion.

einari:
Hope the last commit made it a bit clearer. 
Great to have these kinds of discussions and get the quality of the documentation up - when writing its easy to get trapped and make a lot of assumptions.

Documentation/Architecture/events.md
```diff
@@ -120,16 +120,32 @@ precision on using [primitive types](#Primitive Types).
...
-In addition to events being modelled correctly and the processor doing it right, the system
-must guarantee ordering of events when replaying and also that your [event processor](./event_processor.md)
-can assume that events are being delivered once and once only. This is built into the underlying
-system being used. You as a developer do not have to worry about this.
+This does not mean that you should be keeping full entities on the event - still follow the rule of
+[complexity](#Complexity) as mentioned and remember to model the actual state change.
+Be aware of this and model accordingly.
+
+Important to keep in mind that the snapshot of a state is a series of events, these together form
+the snapshot. This should reduce what you keep on them but as a whole they help guarantee the
```
roarfred: Seems like somethings missing here

roarfred:
Great clarification! Put in a comment about something that might have fallen out, just an incomplete sentence

Documentation/Architecture/events.md
einari: LOL..   Me fix..   Probably a good idea to finish off sentences..

PR #223, opened 2017-10-14
Restructured documentation

roarfred:
Restructured documentation, by introducing the term Processes and changing aggregates to read model


roarfred:
@einari, would it make sense to have the C# command controllers match the processes, so there is a one-to-one relation and a naming convension here?

karolikl:
Waiting to merge pull request until @einari has answered your question.

einari:
That would make perfect sense. It looks good!
 I'm having a discussion with @smithmx to create some clarification around aggregates as we discussed on the Skype call. There might be a concept missing that makes a bit of confusion in this. The controller in this case would be an aggregate root (http://www.dolittle.io/Backend/aggregate_root.html) - and represent the process. We are now discussing whether or not to formalise a difference between the case when it is clearly a process and when it is clearly a noun.  Will return on the topic.

PR #228, opened 2017-10-16
Changing Lat/Long from string to double

roarfred:
Changing data type of Latitude and Longitude from string to double. Seems like there is consensus on double being the most appropriate, due to performance on later geo calculations: https://stackoverflow.com/questions/28068123/double-or-decimal-for-latitude-longitude-values-in-c-sharp

However, there might be two things to rethink:
1) Should we use nullables, allowing a location to be missing?
2) Should we separate a location into a class? (would also allow an empty location)


Source/VolunteerReporting/Events/External/HealthRiskCreated.cs
```diff
@@ -12,5 +12,12 @@ public class HealthRiskCreated : IEvent
         public Guid Id { get; set; }
         public int ReadableId { get; set; }
         public string Name { get; set; }
+        public int? Threshold { get; set; }
+        public string ConfirmedCase { get; set; }
+        public string Note { get; set; }
+        public string SuspectedCase { get; set; }
+        public string ProbableCase { get; set; }
+        public string CommunityCase { get; set; }
```
karolikl: Do we need Threshold, ConfirmedCase, Note, SuspectedCase or ProbableCase in VolunteerReporting?

The KeyMessage we need as this should be included in the messages sent from the system to the data collectors.

roarfred: I also thought about this, but came to the (preliminary) conclusion that the event will however have all attributes from where it was raised, so event if the event processor and read model is just utilizing a very few, the event class itself should be as close to the event as possible. Please chip in and correct me, @einari

roarfred:
Regarding flat and primitive, I see you are correct, thanks! A third possibility can be to separate these into a new events, ____LocationSet and ____LocationCleared. Then we can keep non-nullable types, and still allow for a location to be missing.

I've also seen that use of specific value types (phone number, address, location) over primitive types (int, string etc) is encouraged. So, the question is really more if this applies only to other areas, such as read model and commands, but not to events?
See: [http://www.dolittle.io/Backend/concepts_and_value_objects.html](dolitte.io) 

Again, we might kindly ask @einari for some clarification

einari:
Nullables should never be used. Anything that goes on an event has happened and should never be in doubt. If a value is optional, it should be on an event of its own. There is also a chance the value is represented on a separate command as well - as it is optional. 

For nullables in general; try to stay away from them - it is often a symptom of things being unclear. Something not being set should be drawn back to the domain - it is not a technical decision. It would in fact read much better if you'd have a specifc Concept on for instance the Read Model where the Concept had the correct value that would represent the value not being set. 
Same goes for APIs - they should not return null or nullables - that makes an API unclear as well. 
One of many write-ups on the topic: https://sidburn.github.io/blog/2016/03/20/null-is-evil

The point is; not allowing null or nullables simplifies everything and brings back clarity in your code and a bigger chance of modelling the domain more accurately.

The reason for using primitive types on an event is to make it simpler to version. More complex types, even concepts and values might change in nature and then you'd need to version these - which basically means you'd have to keep all the versions of the complex type around.

roarfred:
I can make some changes to separate the location information into a separate event. How about the read model? Should we also go for separate read models and commands to avoid nullable attributes here too? I can't really see how this plays out, if a user has no location, there will be an empty pointer at one place or the other? Or maybe I have just succumbed to let null be an accepted part of my life :) Please advice!

einari:
Suggestion; what you could do on the read model is to introduce a value object called "Location": 

```csharp
public class Location : Value<Location>
{
    public static readonly Location NotSet;

    static Constructor()
    {
        NotSet = new Location { Longitude = -1, Latitude = -1 };
    }

    public float Longitude { get; set; }
    public float Latitude { get; set; }
}
```

By default you'd set it in the constructor of the Read Model to Location.NotSet.

roarfred:
My other PR #230 caused some conflicts here. Let me clean this up and separate out the Health Risks test data generation before proceeding

Source/VolunteerReporting/Events/External/HealthRiskCreated.cs
karolikl: I've created a separate issue for the discussion on whether all properties should be included on the event or not: #232.

Issue #227, opened 2017-10-16
Generate test data to populate Text Messages that have been received


karolikl:
In VolunteerReporting:

- Generate a test data file for Text Messages, so that we are able to prepoulate the database with some test data.

- In /Source/VolunteerReporting/Web/TestDataGeneratorController.cs, create a new method which emits a TextMessageReceived event per text message in the test data file. For inspiration, take a look at the CreateDataCollectors() where something similar is done for Data Collectors.

PR #230, opened 2017-10-17
Issue #227 generate text message test data

roarfred:
This includes a method to generate some random text messages in a json file, and another method to parse this and emit TextMessageReceived events.

The randomization might be discussed, I wasn't quiet sure how much validity we want in the data... Please advice, and I can adjust accordingly :)


roarfred:
There should now be 90% proper formatted messages, and about 10% of erroneous messages (text or not well formatted). From the proper data, there is a split of 70%/30% of single reports and aggregated reports)

Issue #235, opened 2017-10-18
Create API endpoint: GET api/casereports/


karolikl:
Should GET all casereports from MongoDB.

karolikl:
Note: This issue is dependent on #234.

roarfred:
Just noticed :) I'll include event processing/storage and sample data, unless someone beats me to it...

roarfred:
Ref #234, as there is requested separate storage for anonymous reports, we might need to also have two separate API endpoints for the two types of case reports?

roarfred:
In #239, we are describing that the page listing case reports should also show reports where parsing failed. This means there will have to be event processor, read model and API for the two types of InvalidCaseReports (is it InvalidReport or InvalidCaseReport?)

@alexanderhoset, this should include what you were asking for

Issue #234, opened 2017-10-18
Persist case reports


karolikl:
All case reports created from incoming text messages should be persisted within VolunteerReporting (both anonymous #170 and from known senders #172). 

As we are emitting a CaseReportRecieved event and an AnonymousCaseReportRecieved event, we could create two event processors where we capture our own events and store the contents of the case reports in MongoDB. 
Anonymous and known case reports should be stored separately.

roarfred:
Can/should we rely on our text message processor code to provide test data for case reports too?

Velcrow81:
It says that "Anonymous and known case reports should be stored separately", but from what I can see, they are stored together. The anonymous reports are stored with DataCollectorId = Guid.Empty and the PhoneNumber is not stored. It will be difficult to identify anonymous reports if they all have an empty id and no phone numbers to check against

roarfred:
Agree, my bad! I assume there will be separate read models and logic handling these all the way to the UI too, then?

Velcrow81:
I also assume that😊

20. okt. 2017 07:55 skrev "Roar Fredriksen" <notifications@github.com>:

PR #238, opened 2017-10-18
Issue #234 and #235

roarfred:
This should be a minimal solution for #234, creating a read model and test data for Case Reports, and also for #235, creating a controller for getting all case reports.


roarfred:
I don't think we should remove this, as the CreateAll method would fail. I actually thought of removing the generating method, but decided to keep it until we have found the perfect set of sample data. (Funny what writing the same code too many times, without a duck joke, will make you do)

karolikl:
Two Scottish Ducks were waddling down the road. 
One said "Quack!" 
The other said "I am going as Quack as I can!"

Issue #240, opened 2017-10-19
List of case reports


karolikl:
In Source/VolunteerReporting/Web.Angular project: 
Create UI showing a list of case reports. 

API endpoint: GET /api/casereports

Note: The design has not been implemented yet (#239), so do not spend time optimizing the UI. Simply showing the data is enough in this case.

roarfred:
Fixed by #247

alexanderhoset:
#239 now has a design mockup

woksin:
This is considered done, right?
I suggest closing it.
@karolikl 
@alexanderhoset

PR #243, opened 2017-10-19
Issue #240 make case report list

roarfred:
Not really my strong side, but I got something working by peeking into @anderaus work on the admin project. Please review carefully! There's something about naming that I am not quite confident about, I thought a CamelCaseName should end in a filename of camel-case-name, but could not find this described. Please advice for any adjustments needed!


karolikl:
@anderaus, could you review this?

Source/VolunteerReporting/Web.Angular/src/app/app.module.ts
```diff
@@ -1,24 +1,37 @@
...
 import { BrowserModule } from '@angular/platform-browser';
+import { Routes, RouterModule } from '@angular/router';
 
 import { AppComponent } from './app.component';
-import { CartComponent } from './cart/cart.component';
+import { CaseReportModule } from './casereport/casereport.module';
+import { CoreModule } from './core/core.module';
+
+const routes: Routes = [
+/*
```
anderaus: Commented section can be removed and instead just use the empty array directly on line 31:
`RouterModule.forRoot([]),`

Source/VolunteerReporting/Web.Angular/src/app/casereport/casereportlist/casereportlist.component.html
```diff
@@ -0,0 +1,32 @@
...
+    case report list works!
+</p>
+
+<table>
+    <thead>
+        <tr>
+            <th>Time</th>
+            <th>Data Collector</th>
+            <th>Health Risk</th>
+            <th>Females &gt; 5</th>
```
anderaus: To avoid breaking the heading, you may use no-breaking space instead of breaking space. So something like this:

`<th>Females&nbsp;&gt;&nbsp;5</th>`

Which avoids:

image

Source/VolunteerReporting/Web.Angular/src/app/app.module.ts
roarfred: Fixed, thanks!

Source/VolunteerReporting/Web.Angular/src/app/casereport/casereportlist/casereportlist.component.html
roarfred: Like this table will live for long, lol :) But, thanks, it's fixed

anderaus:
@roarfred not sure I understand the question. Are you talking about when a change has been made to a case report?

roarfred:
No, for the read model being served through the API to the UI. Currently, the Data Collector and the Health Risk is just a guid in the structure being pulled by the service. I assume we'll be better off with a name of the collector and some more details about the health risk, but I see different approaches and can't really tell which one is best practice...

karolikl:
We will need the name and number of the data collector (as the data manager should be able to reply to text messages through the UI in the future), and the name of the health risk. 
Question is, should these be served through a different API endpoint so that the front-end has to do multiple API calls, or should the api/casereports/ endpoint provide this data so the frontend only needs 1 API call?

roarfred:
My gut feeling is that the API should expand these properties, such serving necessary properties in the same call. Otherwise, if you'd like to list 100 events, with 100 unique data collectors and a large data collector database, you'd have to wait for the event list to return and then fetch data collectors. And then, when fetching data collectors, you can either do 100 calls, or collect all unique values to request by an array of ids.

That said, I don't quite trust my instincts here...

karolikl:
I agree, I'm all for limiting the number of calls needed. It also reduces the amount of logic needed in the frontend.

roarfred:
A question here... currently, we're just passing the read model straight from the API to the UI. This means we have a few options:
1) Make an anonymous type (composit from case report, data collector and health risk)
2) Change the read model, so the database also stores the expanded properties
3) Make an alternative composit model based on the three existing read models and return this from the API

Any thoughts? I am leaning towards 3, maybe 2 and rather not 1

PR #247, opened 2017-10-20
Issue240 make case report list

roarfred:
* Separation of anonymous case reports and regular case reports
* a little refactoring for readability

### DIFF
--- a/Source/VolunteerReporting/Web.Angular/src/app/core/anonymouscasereport.service.ts
+++ b/Source/VolunteerReporting/Web.Angular/src/app/core/anonymouscasereport.service.ts
@@ -1,0 +1,21 @@
+import 'rxjs/add/operator/toPromise';
+
+import { Injectable } from '@angular/core';
+import { Headers, Http } from '@angular/http';
+
+import { AnonymousCaseReport } from '../shared/models/anonymouscasereport.model';
+
+@Injectable()
+export class AnonymousCaseReportService {
+    private headers = new Headers({ 'Content-Type': 'application/json' });
+
+    constructor(private http: Http) { }
+
+    getAnonymousCaseReports(): Promise<Array<AnonymousCaseReport>> {
+        return this.http
+            .get('/api/anonymouscasereports', { headers: this.headers })
+            .toPromise()
+            .then((result) => { return result.json(); })
+            .catch((error) => console.error(error));
+    }
+}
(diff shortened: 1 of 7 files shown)


Source/VolunteerReporting/Web.Angular/src/app/core/anonymouscasereport.service.ts
```diff
@@ -0,0 +1,21 @@
...
+
+@Injectable()
+export class AnonymousCaseReportService {
+    private headers = new Headers({ 'Content-Type': 'application/json' });
+
+    constructor(private http: Http) { }
+
+    getAnonymousCaseReports(): Promise<Array<AnonymousCaseReport>> {
+        return this.http
+            .get('/api/anonymouscasereports', { headers: this.headers })
```
karolikl: I'm a bit confused about the path to the endpoint here. When I run this, no data is shown and I get a 404 as it thinks the API is on localhost:4200/api/anonymouscasereports. 
However, if I change the above .get to use localhost:5000/api/anonymouscasereports, it works. Why does this work on your machine and not mine?

anderaus: @karolikl for now, the frontend build output is saved to the wwwroot folder of the backend project, so the intention is that only the backend project should be run (so relative url's work fine). An option is to use a proxy file in the frontend project and run both the backend and frontend projects simultaneously, using the proxy file to route requests from :4200 to :5000. I can create a proxyfile sample if that's preferred?

roarfred: I am using the ng build command (not ng serve) on the angular app, which will package and deploy the application into the wwwroot folder of the dotnet core (Web) application. Then, after running the web project, I can access the UI from port 5000, and service objects will simply use the same port.

karolikl: Aha, now I'm following. What's best practice and will create the least confusion for new contributors? 

I've only followed step 3 here, which runs the frontend on port 5000: https://github.com/IFRCGo/cbs/tree/master/Source/VolunteerReporting 
So maybe step 3 could be simplified then?

anderaus: I prefer the proxy file solution, because that means we can use `ng serve` that will automatically watch for file changes and rebuild&refresh browser on each file change. Makes for an efficient feedback loop when dev'ing. I can setup the Example project with a proxy file. PR coming up!

anderaus: Here you go: https://github.com/IFRCGo/cbs/pull/251

karolikl: As long as it works, I'm happy. Just tried it out and it works like a charm!

PR #251, opened 2017-10-20
Hook up backend and angular properly using proxy file

anderaus:
I removed the use of `ng serve` and set up `npm start` to use a proxy file to forward all requests from the frontend to /api/* to locahost:5000/api/*. Would be great if one of you could confirm that this works well on your machines too.


karolikl:
Nice!

PR #254, opened 2017-10-20
Move event processor from web to read

roarfred:
Just moving some files (event processors and such) from the web project to the read project. While at it, a test project was added for the Read project


roarfred:
Please consider PR #260 first, this one builds upon the same changes

Issue #154, opened 2017-10-01
Zero reporting


aninek:
Data collectors must be able to send a "zero report" when no health risks have been detected for one week. The zero report has the function of telling the system the data collector is still active, but all is fine. 

Format of SMS: 

|Health risk | sex | age |
|--- | --- | --- |
|0 | 0 | 0 |

karolikl:
These should also receive a "Thank you for your report" message in return.

karolikl:
To elaborate, a zero report should result in a standard case report (from either an anonymous or identified data collector), so this issue mainly consists of: 
- Writing test to verify that zero reports are stored as expected.

PR #255, opened 2017-10-20
Make test for zero report

roarfred:
Created a test for #154 
Please note that this PR includes changes in #254, and should be reviewed only after that one

Note: I am out on a limb here... DI, IoC and xunit is all new ground to me. Had to apply a bit of guesswork on the whole mockup thing. The test is working, and testing what it should, but there might be much better ways to do this. Please advice as necessary...


roarfred:
Please consider PR #260 first, this one builds upon the same changes

karolikl:
I'm a bit outdated when it comes to frameworks used for mocking, but we could consider using something like https://fakeiteasy.github.io/. Anyone else with input on this?

einari:
Fake It Easy seems to be quite mature - dating back to 2011. I guess it boils down to personal preference for these kinda things. I must say I really love the syntax for Fake It easy - so much that I'm considering moving away from Moq for [doLittle](https://github.com/dolittle/home) and other project I maintain. 

For your mock objects, you should therefor change to a Mock framework like Fake It easy. 

I'd suggest at the same time we put in a shared `props`file - similar to what I've done [here](https://github.com/dolittle/DotNET.Build/blob/master/MSBuild/specs.props). Then we can gather dependencies we have for typically a test project into one place. And let them reference it like I've done [here](https://github.com/dolittle/DotNET.Fundamentals/blob/master/Specifications/Assemblies/Assemblies.csproj)

PR #261, opened 2017-10-21
Cleaning up builds for all projects

karolikl:
I have cleaned up the appveyor.yml files for all projects and improved the documentation a bit. I have also added build shields for all projects. 

The builds are failing for admin, alerts and user management, so this is something I'll create issues for.


karolikl:
Yup, whitespace... That could have taken a while to figure out, I think I got lucky this time!

PR #266, opened 2017-10-22
Add tests for example project

### LINKED ISSUE #214: Add XUnit tests for Example project

anderaus:
Fixes #214 

I have added two test projects to give examples of how validators and event processors can be tested using FakeItEasy and XUnit.

Only weird thing is I had to target `netcoreapp2.0` instead of `netstandard2.0` to get discovery of tests in Visual Studio working (using `xunit.runner.visualstudio`). Haven't dug into why this happens, but I believe @roarfred has the same issue.

### DIFF
--- a/Source/Example/Catalog/Catalog.sln
+++ b/Source/Example/Catalog/Catalog.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26730.16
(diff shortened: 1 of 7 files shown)


roarfred:
From [xUnit](http://xunit.github.io/docs/getting-started-dotnet-core), they just describe this without much details:
>The result of this project template creates a .NET Standard class library. We actually want a .NET Core class library (since we're writing .NET Core tests), so open the generated .csproj file and edit it(...)

PR #305, opened 2017-11-14
Making sure the event processors are called in all scenarios

### LINKED ISSUE #299: Events are not fired (from infrastructure)
### Description
Just found that the restructuring of the VolunteersReporting project, where the new CaseReporting class of the Domain project is used to raise events causes some problems.

It turns out that none of the events passed to the Apply function of this class ever makes it back to any ICanProcessEvents implementation.

However, events passed to the Apply function of the BaseController class, does find their way through to the corresponding event processors.

The result of this is that no test data for CaseReports will be made. This can be seen by calling the api at /api/testdatagenerator/textmessages

### Environment
* Windows 10
* VS 2017

### Steps to recreate
1. Call the API at /api/testdatagenerator/textmessages
2. See no data generated for case reports

### Current behavior
* Any breakpoint in Process methods of Read.CaseReports.CaseReportEventProcessor will be hit
* No test data is generated for Case Reports

### Expected behavior
* Any breakpoint in Process methods of Read.CaseReports.CaseReportEventProcessor should be hit
* Test data is generated for Case Reports

einari:
Fixes #299

It was using the wrong method - apparently when using `AsyncLocal<>`- it is out of scope when the result is executed. Not entirely sure why that is.

### DIFF
--- a/Source/Infrastructure/AspNet/CommandFilter.cs
+++ b/Source/Infrastructure/AspNet/CommandFilter.cs
@@ -8,20 +8,16 @@ namespace Infrastructure.AspNet
 {
     public class CommandFilter : ActionFilterAttribute
     {
-        static AsyncLocal<bool> _hasEstablishedCommandContext = new AsyncLocal<bool>();
-
         public override void OnActionExecuting(ActionExecutingContext context)
         {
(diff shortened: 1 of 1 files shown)


einari:
Hehe..  True. Good to hear it fixed it for you - sorry it took forever for something this simple. 
Crazy travel month and little energy left in the evenings..

Issue #274, opened 2017-10-26
Get and update automatic reply messages


karolikl:
In Volunteer Reporting, the automatic replies in issue #273  should be editable by the data manager. 

- The messages should have a default/pre-populated value (where should this come from? Admin?).  
- API endpoint needed with a GET and a PUT to retrieve and update the automatic replies.

alexanderhoset:
Point one is the following 
We need two lists of automatic reply messages: 
-System wide fallback
-Project

alexanderhoset:
as I see it the endpoint need:
project id
event 
language code 
message

roarfred:
From #273, I see there are four types of reply messages:
* Unknown sender
* Invalid report (known sender)
* Zero incidents
* One or more incidents

How dynamic do we see these? Do we want to hard-wire the API for these types, open for a more extendable approach using an type/enum, or anything more flexible where rules are defined per project?

I assume a good start can be going for the middle way here, and that "event" as described by @alexanderhoset is the type/enum. As such, we can have the following routes: (also indicated in #275)

GET /api/automaticreply (all messages, for all projects, all languages)
GET /api/automaticreply/{projectId} (all messages for specific project, with defaults from admin/fallback)
PUT /api/automaticreply (only for project specific messages, fallbacks are handled by admin project)

Not sure if the first one is even needed, will the same data manager be working across projects?

alexanderhoset:
My idea is the following(for MVP):
For reporting
1. When a project is created we call getautomaticreply from admin. This adds the fallback messages to the project and the context treat team as project specific messages, . 
2 . All adds or edits are project specific.

for admin 
Get & put automaticreply

karolikl:
Great discussions! One important thing to remember here is that all communication between bounded contexts is done by events, which means that when a project is created from admin, we need the event that is emitted to contain the default reply messages. These will then be stored in VolunteerReporting, and can be overridden by the Data Verifyer in the project. 

I think the endpoints described by @alexanderhoset looks like the right way to go. Of course, the two first are in Admin, so I have created a separate issue for these: #303.

karolikl:
On a second thought: As the automatic replies are only used in VolunteerReporting, maybe the correct solution is that the default messages should also be created/updated there?

alexanderhoset:
its best to split as an admin are not down on project level. it will be the same GUI

alexanderhoset:
I see this  types of reply messages for the MVP:

- Unknown sender
- Valid report
[...]

PR #308, opened 2017-11-14
Handling automatic reply message events from Admin, Read model and API

roarfred:
I know the use of the enum in the event (and then everywhere else) is far from optimal. Can this be a concept? or is there a better suggestion?

**Please check if the PR fulfills these requirements**

- [x] The code builds and tests pass (verified by our automated build checks)
- [x] Commit messages follow this format
```
    Summary of the changes
    - Detail 1
    - Detail 2

    Fixes #issuenumber
```
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Code meets the expectations our engineering guidelines. https://github.com/aspnet/Home/wiki/Engineering-guidelines.

Please review the guidelines for CONTRIBUTING.md for more details.


Source/VolunteerReporting/Web/Models/AutomaticReplyTypeModel.cs
```diff
@@ -0,0 +1,13 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Web.Models
+{
+    public class AutomaticReplyTypeModel
```
einari: Since it sits in Web.Models - does it need to suffixed by Model again?

Source/VolunteerReporting/Web/Models/AutomaticReplyModel.cs
```diff
@@ -0,0 +1,12 @@
+using Events.External;
+
+namespace Web.Controllers
+{
+    public class AutomaticReplyModel
```
einari: Does the technical aspect Model add anything as a suffix?

Source/VolunteerReporting/Web/Controllers/AutomaticRepliesController.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        [HttpGet("{projectId}")]
+        public async Task<IEnumerable<AutomaticReplyModel>> GetAutomaticRepliesForProject(Guid projectId)
+        {
+            var defaults = await _defaultAutomaticReplies.GetAllAsync();
+
+            // TODO: Get replies for project and filter any custom from the default list
+
+            return defaults.Select(c => new AutomaticReplyModel() {
+                Auto = true,
```
einari: Question; What does Auto mean in this context - seeing the model is called AutomaticReply.

Source/VolunteerReporting/Events.External/DefaultAutomaticReplyDefined.cs
```diff
@@ -0,0 +1,15 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Events.External
+{
+    public class DefaultAutomaticReplyDefined : IEvent
+    {
+        public Guid Id { get; set; }
+        public DefaultAutomaticReplyType Type { get; set; }
```
einari: Use an integer as we discussed in another issue that I can't find right now. For make versioning easier.

Source/VolunteerReporting/Web/Models/AutomaticReplyTypeModel.cs
roarfred: Working with both the enum which actually a type and a representation for a view model, this adds something... If we rather use full namespace, or some *using xx as name*, we can skip this. Let me know what you think!

Source/VolunteerReporting/Web/Models/AutomaticReplyModel.cs
roarfred: See my other comment [above](#discussion-diff-151078993R7)

Source/VolunteerReporting/Web/Controllers/AutomaticRepliesController.cs
roarfred: Sorry, this should be called default. Used for distinguishing from the project specific and the global (default) reply messages. Good catch!

Source/VolunteerReporting/Events.External/DefaultAutomaticReplyDefined.cs
roarfred: So, I guess an integer in the event class, and an enum in concept, for readability in code?

Source/VolunteerReporting/Web/Models/AutomaticReplyTypeModel.cs
einari: Not a biggy, I just don't like to have technical concerns added as prefix or suffix to types - the name should clearly say what it is, not where it fits in technically. Is the enum named exactly the same? If so, maybe there is the need to look at naming one of these differently.. ??

Source/VolunteerReporting/Events.External/DefaultAutomaticReplyDefined.cs
einari: Yup - makes sense

Source/VolunteerReporting/Web/Models/AutomaticReplyTypeModel.cs
roarfred: I agree. The enum is actually named DefaultAutomaticReplyType, but I found the Default-prefix here should be removed too. (Types are not specific to the defaults). Would it be better to use a name that more describes the use, such as AutomaticReplyTypeName?

einari: Sure.. Sounds better.

Source/VolunteerReporting/Web/Models/AutomaticReplyModel.cs
roarfred: oh, this was another class than the description of the enum... It's similar in this case, there will be one read model for default automatic reply and one for automatic reply. In the API, you will get a combined list, where you get a default unless it's specified. So, there is a need for wrapping these two models into one, and of course to use a naming that describes this.

Thinking about this, I'll go for AutomaticReplyForProject for the read model, and I can use AutomaticReply just for the API model

roarfred:
Sorry, but I thought this one was merged and committed more stuff to my own master branch. Luckily, it's more code to support #274, so it shouldn't be too bad. Just didn't mean to mess with your review process.

PR #317, opened 2017-11-17
Endpoint for PUT automatic reply

roarfred:
**Please check if the PR fulfills these requirements**

- [x] The code builds and tests pass (verified by our automated build checks)
- [x] Commit messages follow this format
```
    Summary of the changes
    - Detail 1
    - Detail 2

    Fixes #issuenumber
```
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Code meets the expectations our engineering guidelines. https://github.com/aspnet/Home/wiki/Engineering-guidelines.

Please review the guidelines for CONTRIBUTING.md for more details.

PR #318, opened 2017-11-17
Read model for projects

roarfred:
Please note that this PR includes PR #317 so it should only be approved once that one has been approved 

**Please check if the PR fulfills these requirements**

- [x] The code builds and tests pass (verified by our automated build checks)
- [x] Commit messages follow this format
```
    Summary of the changes
    - Detail 1
    - Detail 2

    Fixes #issuenumber
```
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Code meets the expectations our engineering guidelines. https://github.com/aspnet/Home/wiki/Engineering-guidelines.

Please review the guidelines for CONTRIBUTING.md for more details.


Source/VolunteerReporting/Events.External/ProjectUpdated.cs
```diff
@@ -0,0 +1,13 @@
+using doLittle.Events;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Events.External
+{
+    public class ProjectUpdated : IEvent
```
karolikl: If I remember correctly, Admin won't emit one "generic" ProjectUpdated event, but rather specific events such as ProjectInfoChanged or ProjectEventThresholdChanged. 

In this case it looks like we're interested in a ProjectNameChanged, is that right?

roarfred: You are right about this, I didn't check the documentation, just went for the current implementation in the Admin project. Are we OK with driving the definition of the Admin events from the Volunteer Reporting project? If so, I can update this now, but if not we should leave this until work is done inside Admin and the events are finally defined.

karolikl: The event definition should probably come from Admin, so let's keep it for the moment. How do we make sure we remember this change? Add a todo comment?

roarfred: The easy thing would be to make another issue just now to say update the VR project when Admin implements the new event definition, but I soon started to think about a more generic approach... We will be needing a more defined process of handling changes to our cross boundary interfaces, but maybe that's solved by event versioning?

At this stage and maybe later, we'll have to agree on a way where we can:
1) Request spesifics as new or changes to events
2) Get a confirmation on the change (not a working implementation, just a promise this will be implemented) including the exact definition of the event and the version

Now, we can deploy projects individually...

However, at this phase, I'd say we'll make an issue to create the ProjectInfoChange event, and a refernce to another for updating this project. I'll make these

roarfred: Created as #322 and #323

karolikl: We will definitely need a process around this, the issues you have created are a good start :)

Issue #307, opened 2017-11-14
Connecting text messages to a project


roarfred:
In a discussion with @alexanderhoset, we saw the need for having the case reports associated with a project. Pretty obvious, but we couldn't really know for sure where and how an incoming SMS is recognized to be within a specific project. This might also depend on the type of SMS Gateway we are planning on, and  if a gateway will ever serve multiple projects.

We could use both location information, senders phone number or -prefix (nationality or matching data collector), but there might be shortcomings and differences.

While at it, I also wonder how I can determine the preferred language for replying to an unknown number. Might be related to this, but we'll move it to a separate issue if not.

karolikl:
If the same gateway is used for multiple projects, I assume the SMS keyword or the number used will be project specific so that we can route the messages to the correct project. Maybe @ravndal has more input into this?

I find using location data risky as we then assume there will never be several projects within the same area, and this limitation can cause trouble in the future. 

As for the preferred language, could the project have a default language to be used?

ravndal:
Working on a Gateway infrastructure application that will handle all the mapping between incoming messages and projects. 
* Normally we will use keywords as map to one or more projects. 
* In cases where we do not have a keyword, we can map a specific gateway to a project. 

The application will allow multiple keywords and multiple gateways to map to one or more projects.

Issue #152, opened 2017-10-01
Send in report data via SMS


krautmac:
### Description
Data collector should be able to send rapport via standardized SMS. 
Requirements for report SMS to within 160 character limitation to keep costs for the data collector a low as possible

- max length of sms is 841 characters in the gateway

### Report types
Two types of reports: 
a) single case (EVENT,GENDER,AGE) 
b) aggregated (EVENT, # of cases, 4 categories (M<5, M>5, F<5, F>5)

### Things to keep in mind
- todays solution use # and * as a separator
- most newer feature phones have changed the # character button to _shift_ and _T9_ functionality when in SMS view, and the # has to be inserted via special character sheet (=complex)
- A new format with e.g. letters as separators might be more useful and should be user tested in the field

aninek:
Given the difficult use of # in feature phones can we please explore using e.g. * or other as noted above

alexanderhoset:
moved to https://trello.com/c/8xUFQdoK

PR #321, opened 2017-11-19
Inital version of the Gateway implementation

ravndal:
Features:
	- Most of the implementaion should be in place. #307
	- Recive sms #307
	- Send SMS via ISendSMS interface. #152

The gateway we used during the codeathon, is no longer avaialble, so it's not possible to test the complete flow.
Working on a solution to enable testing without an acutal gateway.


ravndal:
NB: The connection string in the code can't really do much harm.

Source/Infrastructure/TextMessaging/TextMessage.cs
```diff
@@ -9,8 +9,9 @@ public class TextMessage
         public string OriginNumber { get; set; }
         public string Message { get; set; }
         public string Keyword { get; set; }
+        public string FullMessage { get; set; }
         public string ReceivedAtGatewayNumber { get; set; }
-        public double Longitude { get; set; }
-        public double Latitude { get; set; }
```
roarfred: Why are we using string representation for the lat/long values?

ravndal: We're not, I'll change it  👍

karolikl:
@einari, can you take a look when you have time?

Issue #108, opened 2017-09-29
Localized Strings


einari:

einari:
I totally agree - moving it...

PR #326, opened 2017-12-03
String localization support

erikly:
Added string localization support in Infrastructure and demo of usage in example project. 

Please note:
- Errors during string resource parsing are not handled. These files should be verified as part of the build process.
- Autofac automatic discovery of ILocalizedStringsManagerFactory didn't work, possibly because of generics. Left a TODO, should be fixed if this is common.

Part of fix to #108, culture in claims is not implemented in this PR


karolikl:
Thanks for the PR, this looks great!

I have asked @einari to do a review as the Infrastructure is his puppy (or so to speak).

Source/Example/Catalog/Web/Localization/ExampleStringManager.cs
```diff
@@ -0,0 +1,14 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) 2017 International Federation of Red Cross. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+ 
+ namespace Web.Localization
+{
+    public class ExampleStringManager
```
einari: Suggest not suffixing this with Manager - as it is not managing anything. So just "ExampleStrings".

einari:
Looks great! 

Suggest we add documentation around this and include best practices around naming and how things should be organised. 
For instance, one of these files per feature - so we don't get dumping grounds for entire bounded contexts. More focused adhering to the principle of high cohesion.

Also want thing that would be interesting moving forward is to allow for having the translation in a place where we don't have to push a new version of the software for translations. Say for instance the JSON files sitting on blob storage. 

Another idea would be for the Angular frontend to by convention be able have a similar infrastructure and just know about the blob path and get it directly from it rather than going through a controller in the backend. We could also proxy generate TypeScript files that one could take a dependency to with the properties and then the infrastructure in the frontend would deal with the similar hookup as done in the backend here.

erikly:
Thanks for the feedback!

Having separate files per feature sounds like a good idea, but it might lead to conflicts with the current JSON structure if multiple features have the same keys. Also the JSONs become the new "dumping ground" if all features are combined here. I suggest creating a set of JSON files per feature with naming "[featureName]-[locale].json" (e.g. "cart-fr.json") to solve this. This might lead to many files, but each file is easy to maintain, and gives fewer conflicts if multiple people work on translation. 
If we choose this solution, I'll change the example to include two different features with separate files and controllers. I like documenting through examples as this makes it very easy to get started, but should there be more documentation elsewhere?

I'll also adjust for future support for remote blob storage by breaking up the retrieval and parsing, so changing to remote storage only involves changing the data source.

The naming is a valid point, I'll fix that too.

To keep a reasonable scope for this PR, I suggest creating new issues for frontend support and remote storage implementation. Also, should I just fix the naming for this PR and solve the other issues separately, or is it better to close this one and solve it all at once?

Any other feedback is also welcome!

einari:
Great, it makes sense to have one JSON file per feature. What if it instead of being named according to the feature - it was located in the folder of where the class representing the strings were and named the same as that class - then you'd have a convention and adhere to the high cohesion principle for that as well.  

For doing something like blob storage, the path within a container could then be following that same structure.

The benefit would be that you'd have a structure, rather than a bunch of files with the structure in the name sitting at root level or somewhere else. This structure could be used for other things we want to keep there as well. 

Totally agree that we can disregard the frontend part for this PR - thats a separate issue and is somewhat of an optimisation. 

On documentation, example is good -thats how I personally like it. But some might want to have it written up and how it works. What do you think @karolikl ?

karolikl:
I think an example showing how it works and how to add new json files should be sufficient. Those who are interested in the implementation details can read the code :)

erikly:
Made changes to address the issues you pointed out, would you like to take another look @einari?

karolikl:
@einari, have you had the chance to look at this?

einari:
Nope.. Not yet..  I'll do it tomorrow @ 15:00. Putting it in my calendar.

Issue #329, opened 2017-12-04
Error: Property 'forkJoin' does not exist on type 'typeof Observable'.


karolikl:
**ERROR in /src/Source/VolunteerReporting/Web.Angular/src/app/core/aggregated-case-report.service.ts (32,27): Property 'forkJoin' does not exist on type 'typeof Observable'.**

I receive this error message when trying to build a Docker image for VolunteerReporting: 
`PS C:\Users\karolikl\OneDrive\Redcross\code\cbs> docker build -t karolikl/cbs-volunteerreporting . -f .\Source\VolunteerReporting\Dockerfile`

It fails when reaching:
+RUN ["ng", "build", "--prod"]

alexanderhoset:
You could try to import this way:
import {Observable} from 'rxjs/Rx';
instead of:
import {Observable} from 'rxjs/Observable';

Just a quick guess

karolikl:
I've made it work by adding: 
import 'rxjs/add/observable/forkJoin'; 

But I have no idea if this is the preferred way of doing it or if there's something better...

alexanderhoset:
Ref comment in code by @roarfred  that  code block needs be be reviewed .... 

@anderaus or @fredrik-lundin is there a better way to combine multiple json calls:
see comment in: /Source/VolunteerReporting/Web.Angular/src/app/core/aggregated-case-report.service.ts

fredrik-lundin:
Regarding the import. If you import it like this:
`import { forkJoin } from "rxjs/observable/forkJoin";`

Then you can use forkJoin without the `Observable` in the code like this:
`forkJoin(...)`

If you import like this:
`import 'rxjs/add/observable/forkJoin';`

Then you use it like this:
`Observable.forkJoin(...)`

Regarding if there is a better way to combine multiple json calls - I don't see the problem with forkJoin in this case. What forkJoin is doing is executing all Observables in parallel, and then emits when all of them have returned something. So from what I can see in the code, it kind of matches with what you are trying to do.

PR #341, opened 2017-12-20
 solves #329 and table fixes for case reports 

alexanderhoset:
#329 and #240 with a status column

### DIFF
--- a/Source/VolunteerReporting/Web.Angular/src/app/core/aggregated-case-report.service.ts
+++ b/Source/VolunteerReporting/Web.Angular/src/app/core/aggregated-case-report.service.ts
@@ -8,6 +8,7 @@ import { Headers, Http } from '@angular/http';
 import { Report } from '../shared/models/report.model';
 import { CaseReport } from '../shared/models/index';
 import { Observable } from 'rxjs/Observable';
+import 'rxjs/add/observable/forkJoin'; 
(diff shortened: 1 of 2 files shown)


Source/VolunteerReporting/Web.Angular/src/app/core/aggregated-case-report.service.ts
```diff
@@ -8,6 +8,7 @@ import { Headers, Http } from '@angular/http';
 import { Report } from '../shared/models/report.model';
 import { CaseReport } from '../shared/models/index';
 import { Observable } from 'rxjs/Observable';
+import 'rxjs/add/observable/forkJoin';
```
roarfred: Might make more sense to just import Observable directly from rxjs, but this can also be fixed later
(This could consolidate a few import statements)
import { Observable } from 'rxjs'

alexanderhoset: Do we close the bug then?

roarfred: I'd say we leave #329 open. This fixes the error, but there's probably a better solution which can clean up these import statements and work as an example on how to do this

PR #342, opened 2017-12-28
Kubernetes deployment and docker-compose for VolunteerReporting backend

karolikl:
**Summary of changes:** 
- Kubernetes deployment of VolunteerReporting backend and mongo db
- docker-compose support in VolunteerReporting

**Note:** 
- I'm not too happy about the speed of building images using docker-compose. It's quite slow and the reason for that is that the entire repo is uploaded to the Docker daemon (when only VolunteerReporting and Infrastructure is needed in this case). I've improved this by adding a .dockerignore, but it's still not good enough. I would prefer to not use root as the build context (as the entire build context is uploaded to the daemon), but I'd like to hear if there are any other opinions on this before I start making changes. 

- As of now, only the backend and db is configured in the Kubernetes deployment and docker-compose. Frontend will follow soon, as well as similar configs for all the other projects.


karolikl:
Update: The previous commits are speeding up the docker-compose builds significantly, so this can be merged. 

Also, I've added frontend to the deployment.

PR #367, opened 2018-01-14
Kubernetes deployment and docker-compose for Admin

karolikl:
Summary of changes:
- Kubernetes deployment of Admin frontend, backend and mongo db
- docker-compose support in Admin

With this change, it should be possible to build and run the project either locally (as before) or by using docker-compose.


alexanderhoset:
In #342  @karolikl  states:
>  I'm not too happy about the speed of building images using docker-compose. It's quite slow and the reason for that is that the entire repo is uploaded to the Docker daemon (when only VolunteerReporting and Infrastructure is needed in this case). I've improved this by adding a .dockerignore, but it's still not good enough. I would prefer to not use root as the build context (as the entire build context is uploaded to the daemon), but I'd like to hear if there are any other opinions on this before I start making changes.

Do we have the same challenge for all deploy setups ?

karolikl:
We don't have the challenge of slow docker-compose builds in this PR, I've managed to speed it up quite a lot. I will commit a fix for #342 this evening.

PR #368, opened 2018-01-15
Kubernetes deployment and docker-compose for User Management

karolikl:
Summary of changes:
- Kubernetes deployment of Admin frontend, backend and mongo db
- docker-compose support in Admin

With this change, it should be possible to build and run the project either locally (as before) or by using docker-compose.

Contains the exact same changes as in #367, but for the User Management project


alexanderhoset:
Do we need it approved by more people to merge @anderaus ?

anderaus:
Sorry, I was meaning to add a comment, @alexanderhoset. I'm not familiar with the docker/kubernetes stuff in the PR, but the Angular bits seem ok. Someone should probably spin this up locally and verify it working, but I don't have the time tonight.

alexanderhoset:
I checked the same, and my docker setup is a bit buggy, so I will wait until we have another review

Issue #350, opened 2018-01-05
User table view


karolikl:
See "Users list" page in cbs/Design/cbs.pdf 

All user types will be displayed in this table, although some users may not have all information.

PR #380, opened 2018-01-16
F add user in modal

cathinkaw:
Started the work to remove Material and also start getting the add user part of the solution in a modal.

Should resolve:
https://github.com/IFRCGo/cbs/issues/352

And is also linked to:
https://github.com/IFRCGo/cbs/issues/350

Sorry about the partly finished rewrite. But I think it is better to get this merged in, than not.

PR #397, opened 2018-01-28
Basic deletion of user in User Management

cathinkaw:
Added simple delete functionality for users in user management. 
Still remainging:
- validation
- confirmation of correct deltion
- reloading list on success

Related to delete link from this issue:
https://github.com/IFRCGo/cbs/issues/350


Source/UserManagement/Web.Angular/src/app/addUser/userForm.component.html
```diff
@@ -0,0 +1,47 @@
...
+        <label for="long">long</label>
+        <input/>
+    </div>
+    <div class="flex-row">            
+        <label for="mobile_number">Mobile phone number(s)</label>
+        <input id="mobile_number" type="tel" formControlName='mobilePhoneNumber' placeholder="Telephone">
+    </div>
+    <div class="flex-row">            
+        <label for="email">Email address</label>
+        <input id="email" type="email" formControlName='email' placeholder="E-mail Address" ngDefaultControl>
```
alexanderhoset: See https://angular.io/api/forms/EmailValidator, then you can skip the regex as we are already using angular/forms for stuff  like this

cathinkaw: This is not part of my change. As this PR was created before https://github.com/IFRCGo/cbs/pull/380 was merged, it looks like there are more changes (this included)

alexanderhoset: Validation need to be fixed later anyway... I will merge this now

PR #461, opened 2018-02-13
Send welcome text after signing up users/collectors

### LINKED ISSUE #375: Send welcome SMS to users after signup
After a user is signed up send them a welcome message defined in admin as a feedback message

This can be done by listening to "our own" user management events: **DataCollectorAdded** and **SaffUserAdded**
(More events can be added as #354 splits SaffUser to more roles )

After message is sent MobilePhoneNumberConfirmed  on the new user is set to true 

notes: 

- Default values for automatic reply messages #303 in admin is not done, so for now set a random message .
- The SMS gateway isn't implemented yet, I suggest we create an abstraction that we later integrate with the SMS gateway.

Davlind:
```
    Summary of the changes
    - Added basic plumbing for sending welcome messages.
    - Admin needs to raise events when welcome messages are changed (including language). Until then we just send a dummy message.
    - SMS needs to pick up the MessageGenerated event and actually send the message.
    - We can discuss my naming of things.....

    Partially fixes #375
```

### DIFF
--- a/Build/build.cake
+++ b/Build/build.cake
@@ -1,117 +1,116 @@
-// //////////////////////////////////////////////////////////////////////
-// // FETCH TOOLS
-// //////////////////////////////////////////////////////////////////////
-// #addin "Cake.Npm"
+//////////////////////////////////////////////////////////////////////
+// FETCH TOOLS
+//////////////////////////////////////////////////////////////////////
+#addin "Cake.Npm"
 
-// //////////////////////////////////////////////////////////////////////
-// // ARGUMENTS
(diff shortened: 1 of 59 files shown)


Davlind:
*Cough* Yea... I should have squashed that...

Davlind:
This was actually made by @einari and pushed to my fork. There was a bug in the framework which prevented you to have multiple event processors for the same event (if I understood it correct). @einari, please correct me if I'm wrong.

karolikl:
Related to #375

PR #466, opened 2018-02-15
Various bug fixes and improvements.

woksin:
@einari Look at https://github.com/IFRCGo/cbs/commit/5e187c86154e712dd9e468459cebc70626ec6cf5, I believe that should be more effective.

### DIFF
--- a/Source/VolunteerReporting/Web/Controllers/CaseReportsController.cs
+++ b/Source/VolunteerReporting/Web/Controllers/CaseReportsController.cs
@@ -35,9 +35,8 @@ namespace Web
             // which was the case when I tested this method using the old method versus fetching the data prior to querying it.
             // In my opinion, the best way to do this is to have a cache-system for these databases (preferably in the classes that deals directly
             // with IMongoDatabase and IMongoCollection 
-            var healthRisks = await _healthRisks.getAllAsync();
-            var dataCollectors = await _dataCollectors.getAllAsync();
-
+            var healthRisks = await _healthRisks.GetAllAsync();
+            var dataCollectors = await _dataCollectors.GetAllAsync();
             // Comment from review; einari - 23rd of October 2017
             // Todo: This is a N+1 query - potentially incredibly slow
             // an optimization would be to get all healthrisks and then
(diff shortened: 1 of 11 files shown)


Source/VolunteerReporting/Web/Controllers/CaseReportsController.cs
```diff
@@ -27,21 +27,32 @@ public CaseReportsController(ICaseReports caseReports, IHealthRisks healthRisks,
...
-        {
+        {   
             var caseReports = await _caseReports.GetAllAsync();
 
+            // Comment from woksin - 15/02-2018
+            // By fetching all risks and dataCollectors to memory we should get reduced latency,
+            // which was the case when I tested this method using the old method versus fetching the data prior to querying it.
+            // In my opinion, the best way to do this is to have a cache-system for these databases (preferably in the classes that deals directly
+            // with IMongoDatabase and IMongoCollection 
+            var healthRisks = await _healthRisks.getAllAsync();
```
Velcrow81: This will work untill data collectors  gets to large, but since they will be stored per national society it might work for the forceable future. Healthrisks should always work :)

woksin: Do you mean it might be a problem considering memory?

einari: There might a need to revisit this whole controller and its models. 
The `CaseReportExpanded` has a very non descriptive name - when looking at it it is really just a projection as well. We could probably do this a bit smarter - **no caching** though. With the events we could have another MongoDB collection that held all this as a projection, rather than having to query all the different ones to get this. 

I do recall us talking a bit about it as a temporary solution...

einari: Btw: Nitpicking; C# convention is UpperCamelCase :) - getAllAsync() -> GetAllAsync()

woksin: @einari Woops.. Java guy here...

woksin: I will fix that asap

Velcrow81: Good point @einari. Did not think about that. Thats kinda the whole idea behind event sourcing and read models :)

woksin: @einari I agree. I'll just let it be as-is for the moment then.

woksin:
Does not look like Travis wants to build..

einari:
Travis build is very unreliable..    Merging now!

Issue #467, opened 2018-02-15
Update the case report model used in the view with a specialized read model that provides everything for the view


Velcrow81:
New read model called CaseReportForListing

woksin:
@Velcrow81 What will this read model be used for exactly? It may be what I'm looking for when optimizing the frontend, look #463.

Velcrow81:
@woksin : This will be the ream model you are looking for. Otherwise I am doing it wrong :)

woksin:
Great!

Velcrow81:
@woksin : Created issue #469 were you could update view to use this new model :)

Issue #184, opened 2017-10-01
When a new data collector is created, check for anonymous case reports linked to the data collectors phone number


karolikl:
When a new data collector is created, he/she might already have sent in reports from an unknown phone number. These have then been saved as "anonymous" case reports. 
So after the data collector has been created, we need to check for anonymous case reports linked to the data collectors phone number. If any anonymous case reports are found, the data collector ID should be added to the report and the CaseReportIdentified (#207) event should be emitted. 

**Further information**
We know that a new data collector has been added by listening to and [processing](https://github.com/IFRCGo/cbs/blob/master/Source/VolunteerReporting/Read/DataCollectorFeatures/DataCollectorEventProcessor.cs) the "DataCollectorAdded" event (#139). 

Path to solution: [https://github.com/IFRCGo/cbs/tree/master/Source/VolunteerReporting](https://github.com/IFRCGo/cbs/tree/master/Source/VolunteerReporting)

roarfred:
I think the correct event to listen for is PhoneNumberAdded. This will be received independently from the DataCollectorAdded event (which doesn't really have a phone number)

karolikl:
Your are absolutely right. Thanks! Update:

**Further information**
We know that a new phone number has been added to a new or existing data collector by listening to and processing the "PhoneNumberAdded" event (#139).

Velcrow81:
Working with events is new to a lot of us and it takes some getting used to. If I process the PhoneNumberAdded event (in Read project?) and find an  "anonymous" case reports that should result in a CaseReportIdentified event beeing emited. However, that means that if we replay the events, another CaseReportIdentified will be emited witch will change the state of the system. My solution would be do change the Process to ProcessOnce once the system supports that (Einar is going to add it), so that this check is only done once per phone number ( never used the word once that many times in a sentance before:)). Is there a better way to do this? Or is that the solution for "every" scenario where we process an event that emits another event?

roarfred:
I don't think that simply emitting an event inside an event processor is a reason to use ProcessOnce. Rather, we should try to keep our event processors idempotent, but only if we can not, we should use the ProcessOnce version. Now guessing, but you might want to both store some data and send an sms from the same event. I guess we should have both a Process (to save data) and a ProcessOnce (to send sms) for the same event. As I understand, you can have multiple event processors, but you can't rely on the order they are called being deterministic.

[...]

Issue #463, opened 2018-02-14
Frontend should fetch a subset of the Case Reports


woksin:
As it is now, the VolunteerReporting/Web.Angular page fetches all Case Reports from the database. This will cause scalabilityproblems in the future when the database grows.
A way to solve this is, as suggested by @karolikl, to let the backend deal with this so that only a maximum of X Case Reports will be fetched from the database.

The backend can deal with this by for example giving the fetch-method some input so that the DB can determine how many Case Reports we want to be showable on the case report list.

woksin:
MVP? @karolikl @Velcrow81

Velcrow81:
Definatly something we need. And case reports are going to be the most numerous entity in our system, so the best place to start :)

PR #468, opened 2018-02-18
Added new read model for case reports view in api. Also added some missing code to closed issue

Velcrow81:
#467: Added a specialized read model that has everything that the case report list view needs. Old controller method is marked as obsolete and view points to this endpoint untill it has been updated to work with the new read model.

Also removed TextMessageId from to read models since it was never set and I dont think it will be used in the future either

Added some questions in code for @einari or @smithmx to look at(and others who like) 

#184: Forgot to identify invalid case reports from unknown data collectors when phone numbers are added. This missing logic has now been added

### DIFF
--- a/Source/VolunteerReporting/Read/CaseReportsForListing/CaseReportForListing.cs
+++ b/Source/VolunteerReporting/Read/CaseReportsForListing/CaseReportForListing.cs
@@ -1,10 +1,5 @@
 using Concepts;
-using Read.CaseReports;
-using Read.DataCollectors;
-using Read.HealthRisks;
 using System;
-using System.Collections.Generic;
-using System.Text;
 
 namespace Read.CaseReportsForListing
 {
@@ -16,10 +11,10 @@ namespace Read.CaseReportsForListing
         public string DataCollectorDisplayName { get; internal set; } = "Unknown"; //QUESTION: Should this be a concept with default value if unknown?
         public HealthRiskId HealthRiskId { get; internal set; }
         public string HealthRisk { get; internal set; } = "Unknown"; //QUESTION: Should this be a concept with default value if unknown?
-        public int? NumberOfFemalesOver5 { get; internal set; } //QUESTION: Should this be a concept with default value if unknown?
-        public int? NumberOfFemalesUnder5 { get; internal set; } //QUESTION: Should this be a concept with default value if unknown?
-        public int? NumberOfMalesOver5 { get; internal set; } //QUESTION: Should this be a concept with default value if unknown?
-        public int? NumberOfMalesUnder5 { get; internal set; } //QUESTION: Should this be a concept with default value if unknown?
+        public int NumberOfFemalesOver5 { get; internal set; }
+        public int NumberOfFemalesUnder5 { get; internal set; } 
+        public int NumberOfMalesOver5 { get; internal set; } 
+        public int NumberOfMalesUnder5 { get; internal set; } 
         public DateTimeOffset Timestamp { get; internal set; }
         public Location Location { get; internal set; }
 
--- a/Source/VolunteerReporting/Read/CaseReportsForListing/CaseReportForListing.cs
+++ b/Source/VolunteerReporting/Read/CaseReportsForListing/CaseReportForListing.cs
@@ -1,0 +1,34 @@
+using Concepts;
+using Read.CaseReports;
+using Read.DataCollectors;
+using Read.HealthRisks;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Read.CaseReportsForListing
+{
+    public class CaseReportForListing
+    {
+        public Guid Id { get; private set; }
+        public CaseReportStatus Status { get; internal set; }
+        public DataCollectorId DataCollectorId { get; internal set; }
+        public string DataCollectorDisplayName { get; internal set; } = "Unknown"; //QUESTION: Should this be a concept with default value if unknown?
+        public HealthRiskId HealthRiskId { get; internal set; }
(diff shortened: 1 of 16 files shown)


Source/VolunteerReporting/Read/CaseReportsForListing/CaseReportForListing.cs
```diff
@@ -0,0 +1,34 @@
...
+    public class CaseReportForListing
+    {
+        public Guid Id { get; private set; }
+        public CaseReportStatus Status { get; internal set; }
+        public DataCollectorId DataCollectorId { get; internal set; }
+        public string DataCollectorDisplayName { get; internal set; } = "Unknown"; //QUESTION: Should this be a concept with default value if unknown?
+        public HealthRiskId HealthRiskId { get; internal set; }
+        public string HealthRisk { get; internal set; } = "Unknown"; //QUESTION: Should this be a concept with default value if unknown?
+        public int? NumberOfFemalesOver5 { get; internal set; } //QUESTION: Should this be a concept with default value if unknown?
+        public int? NumberOfFemalesUnder5 { get; internal set; } //QUESTION: Should this be a concept with default value if unknown?
```
woksin: Is there a specific reason why these properties can be nullable? Isn't it more logical to have them not be nullable, but 0 instead if that value have no meaning?

woksin:
This makes it much nicer to work with the reports from a frontend perspective, which is nice! Fetching data from the Web.App would be much smoother and more effective.
But I think it adds another layer of complexity,

Source/VolunteerReporting/Read/CaseReportsForListing/CaseReportForListing.cs
Velcrow81: I made them nullable to make it more visible that the case report was not parsed correctly, so we do not know what the number should be. You also have something called zero reports, but since that would have status success, it would be easy to differentiate between them. I can change this and update if I get any better tips later :)

Velcrow81:
Thank you for the review :)

 It will get huge so it will need some sort of paging later. I think this is something that @einari will introduce later though doLittle?

It will need to revisit the read model many times (since we do not use many of them in views yet and stuff changes), but the whole idea is that you should not be afraid of repeating as this often removes dependency complexity. The read models are supposed to be specialized "caches" that gives exactly what is needed to the one situation it is used for. There might be exceptions, i.e. massive data sizes, but in this project I think it will work. The point of the read model is to provide fast and easy data were it is needed, and since the application is event sourced, it is easy to update (many places if needed) as data change. The read model will change and improve as we get to know what is needed.

Source/VolunteerReporting/Read/CaseReportsForListing/ICaseReportsForListing.cs
```diff
@@ -0,0 +1,13 @@
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Read.CaseReportsForListing
+{
+    public interface ICaseReportsForListing
+    {
+        Task<IEnumerable<CaseReportForListing>> GetAllAsync();
```
woksin: Would be nice to have a Task<IEnumerable<CaseReportForListing>> GetLimit(int limit) which fetches the limit last or first items. #463

Source/VolunteerReporting/Read/CaseReportsForListing/CaseReportForListing.cs
```diff
@@ -0,0 +1,29 @@
...
+using System;
+
+namespace Read.CaseReportsForListing
+{
+    public class CaseReportForListing
+    {
+        public Guid Id { get; private set; }
+        public CaseReportStatus Status { get; internal set; }
+        public DataCollectorId DataCollectorId { get; internal set; }
+        public string DataCollectorDisplayName { get; internal set; } = "Unknown"; //QUESTION: Should this be a concept with default value if unknown?
```
karolikl: How will this work with localization in the future? We can either localize the default value in the read model or should frontend be in charge of retrieving the correct value based on the users language?

Source/VolunteerReporting/Read/CaseReportsForListing/CaseReportForListing.cs
```diff
@@ -0,0 +1,29 @@
...
+        public HealthRiskId HealthRiskId { get; internal set; }
+        public string HealthRisk { get; internal set; } = "Unknown"; //QUESTION: Should this be a concept with default value if unknown?
+        public int NumberOfFemalesOver5 { get; internal set; }
+        public int NumberOfFemalesUnder5 { get; internal set; } 
+        public int NumberOfMalesOver5 { get; internal set; } 
+        public int NumberOfMalesUnder5 { get; internal set; } 
+        public DateTimeOffset Timestamp { get; internal set; }
+        public Location Location { get; internal set; }
+
+        //QUESTION: Should we also add text messages that could not be parsed with the parsing error messages? Or is it enough with that status, and another read model should be used in a different view were invalid case reports can be fixed
```
karolikl: For the MVP at least, I believe it will be enough with a status.

Source/VolunteerReporting/Read/CaseReportsForListing/CaseReportForListing.cs
Velcrow81: Good point. What you said 😊 I have no solution atm

Source/VolunteerReporting/Read/CaseReportsForListing/CaseReportForListing.cs
Velcrow81: I agree 😊

Issue #469, opened 2018-02-18
Update case report view to use the new specialized read model


Velcrow81:
The read model does not contain details on failed or unknown case reports, just a status (according to the mockup). It is assumed that this information will be shown in other views were i.e. users can fix messages that failed parsing. Is this correct @karolikl and @alexanderhoset ?

Velcrow81:
Related to Update the case report model used in the view with a specialized read model that provides everything for the view #467

woksin:
I will take a look at this it today.

woksin:
@Velcrow81 Close this? #472

PR #472, opened 2018-02-20
#469 Updated the view for Case Reportings

woksin:
Updated the VolunteerReporting view to use the new CaseReportForListing model.
Fixed some errors in Read/CaseReportForListing giving CaseReportForListing a Message string so that the view can show the TextMessage.Message of an CaseReport which is invalid.
Also made sure that the GenerateTestDataController deletes the new CaseReportsForListing Collection.


woksin:
Should solve issue #469, @Velcrow81 ?
I did not to really extensive testing on the frontend part, sorting and filtering. But it should work.

PR #476, opened 2018-02-21
If HealthRiskId is Guid.Empty, report an invalid message

### LINKED ISSUE #473: CaseReportForListing collection contains 99/100 reports
It seems like not all of the reports are put into the CaseReportForListing collection.
Here are screenshots I took when I was debugging using the swagger Controller interface and using only the GetAllAsync methods in each respective collection: CaseReport-collection CaseReportFromUnknown..-collection, ...

If you want to replicate this try putting a breakpoint infront of each collection's GetAllAsync-method.

karolikl:
Checking if a healthrisk is not found, and if so reporting an invalid message
Fixes #473

### DIFF
--- a/Source/VolunteerReporting/TextMessaging/TextMessageProcessor.cs
+++ b/Source/VolunteerReporting/TextMessaging/TextMessageProcessor.cs
@@ -54,8 +54,13 @@ namespace TextMessaging
                 return;
             }
 
-            // Todo: Should we have a validation check if we actually get a health risk id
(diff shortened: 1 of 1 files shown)


Velcrow81:
@karolikl : Thanks for the bug fixing :)

I think we should ensure that this situation is documented in the error message as well, so its easier to understand why the message failed. Will you do it or should I create a new issue for it?

karolikl:
I'll do it :) I'd like to do some refactoring as well to include the health risk check as part of the parsing.

Velcrow81:
@karolikl: I did it in the last PR(error message). Have a look and see if it makes sense 😊

PR #477, opened 2018-02-22
#463 Frontend can fetch a subset of CaseReportsForListing.

woksin:
Added a method in CaseReportsForListing that fetches a maximum amount, limit, of reports from its collection either from the front or back specified by a boolean argument.

Re-added the obsolete method from CaseReportController that retrieves the reports from CaseReports collection. It's nice to have for testing.

The frontend stuff with TS and Angular frustrates me, my efficiency is drastically reduced when working with the frontend stuff. We need someone to look at the frontend issues for VolunteerReporting.

### DIFF
--- a/Source/UserManagement/Web/Controllers/UserController.cs
+++ b/Source/UserManagement/Web/Controllers/UserController.cs
@@ -14,7 +14,7 @@ using Infrastructure.AspNet;
 
 namespace Web
 {
-    [Route("api/usermanagement/")]
+    [Route("api/staffuser")]
     public class UserController : BaseController
     {
         readonly IUsers _users;
@@ -28,7 +28,7 @@ namespace Web
             _logger = logger;
         }
 
-        [HttpPost("staffuser")]
+        [HttpPost]
         public void Add([FromBody] AddStaffUser command)
         {
             var id = Guid.NewGuid();
@@ -53,8 +53,25 @@ namespace Web
             Console.WriteLine("in staffusers");
             var users = _users.GetAllStaffUsers();
             return users;
-        }        
+        }
 
+        [HttpPost("update/{id}")]
+        public void Update([FromBody] AddStaffUser command, Guid id)
+        {
+            Apply(id, new StaffUserAdded
+            {
+                Id = id,
+                FullName = command.FullName,
+                DisplayName = command.DisplayName,
+                Age = command.Age,
+                Sex = command.Sex,
+                Location = command.Location,
+                NationalSociety = command.NationalSociety,
+                PreferredLanguage = command.PreferredLanguage,
+                MobilePhoneNumber = command.MobilePhoneNumber,
+                Email = command.Email
+            });
+        }
         [HttpDelete("delete/{id}")]
         public void Delete(Guid id)
         {
--- a/Source/VolunteerReporting/Read/CaseReportsForListing/CaseReportsForListing.cs
+++ b/Source/VolunteerReporting/Read/CaseReportsForListing/CaseReportsForListing.cs
@@ -1,9 +1,15 @@
 using MongoDB.Driver;
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
+using System.Reflection;
 using System.Text;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization;
+using Newtonsoft.Json.Converters;
 
 namespace Read.CaseReportsForListing
 {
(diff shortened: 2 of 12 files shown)


Source/VolunteerReporting/Read/Read.csproj
```diff
@@ -13,5 +13,10 @@
     <PackageReference Include="doLittle.Events.Processing" Version="2.0.0-alpha2.194" />
     <PackageReference Include="doLittle.Runtime.Events" Version="2.0.0-alpha2.70" />
   </ItemGroup>
+  <ItemGroup>
+    <Reference Include="Microsoft.AspNetCore.Mvc.Core">
+      <HintPath>..\..\..\..\..\..\..\..\Program Files\dotnet\sdk\NuGetFallbackFolder\microsoft.aspnetcore.mvc.core\2.0.2\lib\netstandard2.0\Microsoft.AspNetCore.Mvc.Core.dll</HintPath>
```
woksin: Weird, the Read project should not have use of this.

Source/VolunteerReporting/Web.Angular/src/app/case-report/case-report-list/case-report-list.component.ts
```diff
@@ -33,10 +34,19 @@ export class CaseReportListComponent implements OnInit {
...
+        "timestamp", "location"
+    ];
+
     constructor(
         private caseReportService: AggregatedCaseReportService,
-        private service: ReportService
-    ) { this.maxReports = 1000 }
+        private service: ReportService,
+        private caseReportExporter: CaseReportExporter
+    ) { this.maxReports = 10 }
```
woksin: Should change this to a bigger number, 100 for example.

Source/UserManagement/Web/Controllers/UserController.cs
```diff
@@ -14,7 +14,7 @@
 
 namespace Web
 {
-    [Route("api/usermanagement/")]
+    [Route("api/staffuser")]
```
Velcrow81: Restfull api naming is typically plural, so staffusers is more the norm :)

Source/VolunteerReporting/Read/CaseReportsForListing/CaseReportsForListing.cs
```diff
@@ -26,6 +33,26 @@ public CaseReportsForListing(IMongoDatabase database)
             return await list.ToListAsync();
         }
 
+        public async Task<IEnumerable<CaseReportForListing>> GetLimitAsync(int limit, Boolean last)
+        {
+            
+            var filter = Builders<CaseReportForListing>.Filter.Empty;
+
+            // TODO: Can we assume that this explicit conversion is safe?
```
Velcrow81: Don't know

Source/VolunteerReporting/Read/CaseReportsForListing/CaseReportsForListing.cs
```diff
@@ -26,6 +33,26 @@ public CaseReportsForListing(IMongoDatabase database)
...
+            var totalDocuments = (int)await _collection.CountAsync(filter);
+
+            limit = (limit > totalDocuments) ? totalDocuments : limit;
+
+            var options = new FindOptions<CaseReportForListing>()
+            {
+                Skip = last ? totalDocuments - limit : 0,
```
Velcrow81: Is this correct? If you don't want the first 100, but the following 100 in a collection of 1000, you enter 100 and last = true. That means 1000 - 100 = 900, i.e. you skip the first 900, not just 100. If you enter 900, then you skip the first 100, but it sounds strange with a limit of 900 to get 100. Maybe limit is not right word? Am I missing something?

Source/VolunteerReporting/Web.Angular/src/app/case-report/case-report-list/case-report-list.component.ts
```diff
@@ -33,10 +34,19 @@ export class CaseReportListComponent implements OnInit {
 
     basicFilter: string = "all";
     
+    fields: Array<string> = [
+        "id", "status", "dataCollectorId", "dataCollectorDisplayName",
+        "healthRiskId", "healthRisk", "message",
+        "numberOfFemalesOver5", "numberOfFemalesUnder5",
```
Velcrow81: I have just renamed the age properties. Ensure that these are correct after merge with my PR :)

Source/UserManagement/Web/Controllers/UserController.cs
woksin: Okay, I will keep that in mind then.

Source/VolunteerReporting/Read/CaseReportsForListing/CaseReportsForListing.cs
woksin: Well, limit is just the max amount of reports you want to fetch from the collection. And depending on what you want you can retrieve the limit last elements or the limit first

woksin: In my implementation here there is not really room for getting the "in between" documents...
It all very much depends on what you want the system to do really.

woksin: Limit= 900 and last = true would fetch the last 900 documents, skipping the first Collection.Size - 900 documents.

Source/VolunteerReporting/Web.Angular/src/app/case-report/case-report-list/case-report-list.component.ts
woksin: Will do

Issue #471, opened 2018-02-19
Update staff user


karolikl:
Add endpoint for updating staff users in UserController.cs.

Please also update the route of the UserController to to api/
staffuser instead of api/usermanagement/staffuser.

woksin:
Dependent on #354, @karolikl ?

karolikl:
Correct :)

woksin:
Similar to #499

karolikl:
@woksin, can this be closed?

woksin:
@karolikl very soon, I'll implement the rest of the update functionality before monday

Issue #470, opened 2018-02-19
Update data collector


karolikl:
Add endpoint for updating data collectors in DataCollectorsController.cs. 

If a phone number is added, the PhoneNumberAddedToDataCollector event (#219) must be emitted. 

If a phone number is removed, the PhoneNumberRemovedFromDataCollector event (#220) must be emitted.

woksin:
Dependent on #499 to figure out a system for sending update events

Issue #354, opened 2018-01-05
Clean up StaffUser (read models, commands, events)


karolikl:
**Update after codeathon2018:**
We have started the work of splitting up the "staffUser" into the several roles described below.
Decisions made:
- We're keeping it down to one endpoint for all users, with the same command for all users.
- The validator for the addUser-command validates based on Role and validated which fields are relevant for which roles.
- Controller will send command into Aggregate Root
- Aggregate Root will split up into several separate events for the different users based on Role (work begun by @smithmx )

**Original description:**
The StaffUser read model (\cbs\Source\UserManagement\Read\StaffUser.cs) needs to be split up into separate user types: 

All roles have the following properties: 
- Full name (Mandatory, editable)
- Display Name (Mandatory, editable)
- Email address (Mandatory, editable)

**Admin**
- See above

**System Configurator**
- See above, plus
- Birth year 
- Sex (Optional, non-editable)
- National Society (Mandatory, editable) @RFMoore need to include IFRC GVA and CO and ICRC GVA CO
- Preferred language (Mandatory, editable)
- Mobile phone number(s) (Mandatory, editable)
- MobilePhoneNumberConfirmed boolean (always true for MVP) (automated)
- Assigned National Societies (Mandatory, editable)

**Data Coordinator**
- See above, plus
- Age (Optional, _may need to move to birthyear as to change over time_)
- Sex (Optional, non-editable)
- National Society (Mandatory, editable) OR IFRC or ICRC GVA, region or CO
- Preferred language (Mandatory, editable)
- Mobile phone number(s) (Mandatory, editable)
- MobilePhoneNumberConfirmed boolean (always true for MVP) (automated)
- Assigned National Societies (Mandatory, editable)

**Data Owner**
- See above, plus
- Birthyear (Optional, editable)
- Sex (Optional, non-editable)
- National Society (Mandatory, editable)
- Preferred language (Mandatory, editable)
- Mobile phone number(s) (Mandatory, editable)
- MobilePhoneNumberConfirmed boolean(always true for MVP) (automated)
- Position (free text, position within the NS) (Mandatory, editable)
- Duty station

**Data Verifier**
- See above, plus
- Birthyear (Mandatory)
- Sex (Mandatory, non-editable)
- National Society (Mandatory, editable)
- Preferred language (Mandatory, editable)
- GPS location / Location (area) (Mandatory, editable)
- Mobile phone number(s) (Mandatory, editable)
[...]

PR #482, opened 2018-02-24
#354 Major cleanup to Read models, commands and events

woksin:
edit: Take a look at my github repo to have a better look at the structural changes https://github.com/woksin/cbs/tree/master/Source/UserManagement
Sorry for big push...

I did some major rework on the project, mostly in terms of restructuring.
Split the Add/Remove PhoneNumber Command into two, one for Datacollector Staffusers.

Split Users into multiple Read models as per defined by #354. 

Added a TestDataGeneratorController for generating a dataset for testing, importing testset as commands and sending them to the appropriate commandhandlers.

Command => Handling => Event processing Works now and is (hopefully) implemented the way the doLittle guys want(?)

I commented out the code inside the Domain.Specs project since it would take some time to make it compatible with the new commands I made (and I frankly don't actually know what's going on with the code there). Someone should perhaps have a look at it.

I'm open for suggestions on improvement and domain specific stuff.
I kind of made improvements where I saw fit, I'm happy with the current structure of the project, but it's not perfect so come with suggestions on improvements. 

Work remaining:
Add testdatagenerator for trying out the Add/Remove PhoneNumber commands.
Take a look at the GreetingHistory read model.
Add update endpoints, commands and events for updating DataCollectors and StaffUsers.
[Take a look at my branch for adding update endpoints?](https://github.com/woksin/cbs/tree/%23471-%23470/Source/UserManagement)
#470 #471


woksin:
Important to note. I recognized a problem with only using async methods when updating/adding database entries. If the system receives a command that splits up into multiple events that gets processed into queries that fetches and updates on the same database entry we can get synchronization problems.

I noticed that when testing out a branch that implements the update functionality on datacollector. I solved the problem there by switching to the synchronous methods provided by the collection interfaces.

I will fix thisproblem across the whole project in a later pull request.

Source/UserManagement/Domain/DataCollectors/Validators/AddPhoneNumberToDataCollectorValidator.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace Domain.DataCollectors.Validators
+{
+    public class AddPhoneNumberToDataCollectorValidator : AbstractValidator<AddPhoneNumberToDataCollector>
+    {
+        public AddPhoneNumberToDataCollectorValidator()
+        {
+            CascadeMode = CascadeMode.StopOnFirstFailure;
+
+            RuleFor(_ => _.DataCollectorId)
+                .NotEmpty().WithMessage("DataColelctorId cannot be empty")
```
karolikl: Small typo: DataColelctorId -> DataCollectorId

Source/UserManagement/Domain/DataCollectors/Validators/RemovePhoneNumberFromDataCollectorValidator.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace Domain.DataCollectors.Validators
+{
+    public class RemovePhoneNumberFromDataCollectorValidator : AbstractValidator<RemovePhoneNumberFromDataCollector>
+    {
+        public RemovePhoneNumberFromDataCollectorValidator()
+        {
+            CascadeMode = CascadeMode.StopOnFirstFailure;
+
+            RuleFor(_ => _.DataCollectorId)
+                .NotEmpty().WithMessage("DataColelctorId cannot be empty")
```
karolikl: Small typo: DataColelctorId -> DataCollectorId

Source/UserManagement/Events/DataCollector/PhoneNumberAddedToDataCollector.cs
```diff
@@ -0,0 +1,20 @@
+using doLittle.Events;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Events.DataCollector
+{
+    public class PhoneNumberAddedToDataCollector : IEvent
+    {
+        // QUESTION: Needed? Is this EventSourceId? If so, then all events should have this? public Guid Id { get; set; }
```
karolikl: I can't see how this is needed, events should only provide information relevant to the event. Comments, @einari?

woksin:
Thanks for the review, @karolikl . I'll look into the things you commented on.
I will also push to this PR the changes that reflects the new requirements on #354.

@karolikl I like the renaming DataCollector.cs => DataCollectorManagement.cs because:
1) In the Read project there already exists a DataCollector.cs 
2) It, in my opinion, signifies a class that handles management of the class it operates on.
Though I could agree with a name like DataCollectorAggregateRoot or DataCollectorRoot :)

Source/UserManagement/Events/DataCollector/PhoneNumberAddedToDataCollector.cs
woksin: @einari for further information, the property in question is a
public Guid id

Source/UserManagement/Domain/DataCollectors/Commands/AddDataCollector.cs
```diff
@@ -7,20 +7,20 @@
...
 using doLittle.Commands;
 
-namespace Domain
+namespace Domain.DataCollectors.Commands
 {
     public class AddDataCollector : ICommand
     {
         //TODO: Update these properties to reflect what is needed for event. Remove PhoneNumber
-        public Guid Id { get; set; }
+      //  public Guid Id { get; set; } // TODO: Question: Commands should not contain the Id of the result object?
```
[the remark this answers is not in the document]
woksin: @karolikl I see. Makes sense. But still, if we can assume that the commands are supplied only through the Web Controllers, then the idea would be to relieve the frontend (angular frontend or in general the incomming HTTP request to the api provided by the controller) from the job of creating the Guid and instead make the controller that recieves the request handle creation of the Guid?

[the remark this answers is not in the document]
woksin: I see that this was the way it was done previously... My bad, I will change it back to the method i mentioned above.

Source/UserManagement/Events/DataCollector/PhoneNumberAddedToDataCollector.cs
woksin: @karolikl May have been something that I forgot to remove after a copy-paste or something also

Velcrow81:
@smithmx: Could you have a look at the proposed folder structure in domain and evaluate og this should be part of the"golden path" that should be used as an example by others.

woksin:
The question is though if the structure and flow of commands that is proposed here is the right way, or atleast on that path.

Source/UserManagement/Domain/StaffUser/StaffUser.cs
```diff
@@ -0,0 +1,96 @@
...
+using doLittle.Runtime.Events;
+using Events.StaffUser;
+namespace Domain.StaffUser
+{
+    public class StaffUser : AggregateRoot
+    {
+        protected StaffUser(EventSourceId id) : base(id) {}
+
+        //TODO: I have no idea, not even sure what AggregateRoot means. Someone should look over this.
+        public void AddStaffUser(AddStaffUser command)
```
smithmx: Don't add commands to the Aggregate Root.  The Aggregate Root should expose methods with parameters that are needed.  It's the role of the Command Handler to call the correct method on the aggregate root and to pass the appropriate parameters.

The AddStaffUser command also seems quite "CRUD"y.  A parameter that distinguishes between different types is usually a code smell.  AddAdmin, AddDataConsumer, AddDataCoordinator, etc. are much better commands.  They express the intent without having to interrogate the values of the command.

Source/UserManagement/Domain/StaffUser/StaffUser.cs
```diff
@@ -0,0 +1,96 @@
...
+                case Role.SystemCoordinator:
+                    HandleSystemCoordinator(command);
+                    break;
+            }
+        }
+
+        private void HandleAdmin(AddStaffUser command)
+        {
+            Apply(new AdminAdded(
+                Guid.NewGuid(), command.FullName,
```
smithmx: The Id of the Admin to be created should be on the Command.   We don't return any state from the CommandHandler so if we wanted to navigate to the newly created Admin we wouldn't be able to.  If the Guid is generated on the front end as part of the Command, we could then navigate to that Guid given a successful CommandResult.

Where a legacy system creates an auto-generated incremental key (think IDENTITY in Sql Server) for example, I tend to add Guid as a surrogate key and use that for the initial navigation and then the system generated key from then on.

Source/UserManagement/Events/StaffUser/AdminAdded.cs
```diff
@@ -0,0 +1,24 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Events.StaffUser
+{
+    class AdminAdded
```
smithmx: Yes!  Creating multiple events, one for each scenario, is correct.

Source/UserManagement/Web/Controllers/StaffUserController.cs
```diff
@@ -0,0 +1,93 @@
...
+using Microsoft.Extensions.Logging;
+using doLittle.Events;
+using Read;
+using Infrastructure.AspNet;
+using Read.StaffUsers;
+
+namespace Web
+{
+    [Route("api/staffusers")]
+    public class StaffUserController : BaseController
```
smithmx: This is not how it should be done.  The controller should just delegate to the CommandHandler.

Source/UserManagement/Concepts/PhoneNumber.cs
```diff
@@ -0,0 +1,36 @@
...
+        public PhoneNumber(string value)
+        {
+            Value = value;
+        }
+
+        
+        //TODO: This value should default to false after the MVP and when there is logic for phone number confirmation
+        public bool Confirmed { get; private set; } = true;
+
+        public override bool Equals(object obj)
```
smithmx: There's no need to provide overrides for Equals and GetHashCode.  This is what Value<T> does.  It applies Value equality semantics (comparing on type and value of properties).

Source/UserManagement/Concepts/Role.cs
```diff
@@ -11,6 +11,14 @@ public enum Role
 
     public static class RoleExtensions
     {
+        public static bool RequiresAgeAndSex(this Role r) =>
```
smithmx: This is purely a style thing, but I prefer to make the enum a class when it starts to have extra info or behaviour.  In the interests of full disclosure, I should note that I'm not a huge fan of enums in general. :-)

Source/UserManagement/Domain/DataCollectors/AggregateRoots/DataCollectorManagement.cs
```diff
@@ -0,0 +1,61 @@
+using System;
+using doLittle.Domain;
+using Domain.DataCollectors.Commands;
+using Events.DataCollector;
+
+namespace Domain.DataCollectors.AggregateRoots
```
smithmx: There shouldn't be folders based on technical role.  That is, no AggregateRoots, Validators, CommandHandlers or Commands.  Things should be grouped by Feature.  Where there is too many artifacts, this might indicate that you are missing sub features.

Source/UserManagement/Domain/DataCollectors/CommandHandlers/DataCollectorCommandHandler.cs
```diff
@@ -0,0 +1,43 @@
...
+            _repository = repository;
+        }
+
+        public void Handle(AddDataCollector command)
+        {
+            var root = _repository.Get(command.DataCollectorId);
+            root.AddDataCollector(command);
+        }
+
+        public void Handle(AddPhoneNumberToDataCollector command)
```
smithmx: Again, no commands passed to the aggregate root.

root.AddPhoneNumber(command.PhoneNumber)

I am also a bit suspicious of the Add* and Remove* names as it feels like CRUD.  I don't know the domain well enough to say if this is correct or not though.

Source/UserManagement/Domain/DataCollectors/CommandHandlers/IDataCollectorCommandHandler.cs
```diff
@@ -0,0 +1,12 @@
+using doLittle.Runtime.Commands;
+using Domain.DataCollectors.Commands;
+
+namespace Domain.DataCollectors.CommandHandlers
+{
+    public interface IDataCollectorCommandHandler : ICanHandleCommands
```
smithmx: I don't think there is any benefit in adding an interface for IDataCollectorCommandHandler.

I might group Handle methods in a particular class for convenience (where they are related to the same Aggregate Root and have the same dependencies) but since the infrastructure / framework has the responsibility for hooking up commands to the handlers, this is not something I would test / mock out.

Source/UserManagement/Domain/DataCollectors/Commands/AddDataCollector.cs
```diff
@@ -4,23 +4,24 @@
...
         public string FullName { get; set; }
         public string DisplayName { get; set; }
         public int YearOfBirth { get; set; }
-        public Sex Sex { get; set; } //Do we need Transgender / Other?
+        public Sex Sex { get; set; }
         public Guid NationalSociety { get; set; }
         public Language PreferredLanguage { get; set; }
         public Location GpsLocation { get; set; }
-        public string MobilePhoneNumber { get; set; }
+        public List<string> MobilePhoneNumbers { get; set; }
```
smithmx: I'd use string[]

Source/UserManagement/Domain/DataCollectors/Validators/AddPhoneNumberToDataCollectorValidator.cs
```diff
@@ -0,0 +1,23 @@
...
+using Domain.DataCollectors.Commands;
+using FluentValidation;
+
+namespace Domain.DataCollectors.Validators
+{
+    public class AddPhoneNumberToDataCollectorValidator : AbstractValidator<AddPhoneNumberToDataCollector>
+    {
+        public AddPhoneNumberToDataCollectorValidator()
+        {
+            CascadeMode = CascadeMode.StopOnFirstFailure;
```
smithmx: This is a global validation mode.  I prefer to get all the issues with the command so that these can be displayed to the user.  

Sometimes you have multiple rules on a single property that depends on it not being null for example.  Then you can set the Cascade mode on the property and have the null check first.

Source/UserManagement/Domain/MessageGenerator/MessageGenerator.cs
```diff
@@ -1,10 +1,8 @@
 using doLittle.Domain;
 using Events;
 using System;
-using System.Collections.Generic;
-using System.Text;
 
-namespace Domain.MessageGenerators
+namespace Domain.MessageGenerator
```
smithmx: I'm not sure MessageGenerator is an Aggregate Root.  I think it might be needed just now for technical reasons (I don't think policies can publish events).

Source/UserManagement/Domain/StaffUser/AggregateRoots/StaffUserManagement.cs
```diff
@@ -0,0 +1,161 @@
...
+
+namespace Domain.StaffUser.AggregateRoots
+{
+    public class StaffUserManagement : AggregateRoot
+    {
+        public StaffUserManagement(Guid id) : base(id)
+        {
+        }
+        
+        public void AddStaffUser(AddStaffUser command)
```
smithmx: This should be mutiple commands.  Put the role into the command name and let the front end send the correct command based on what role is selected.  Pretty much you have the command names, they are the methods minus the Handle prefix. 

In general terms, a property that indicates a type is a code smell in DDD.

Also, commands shouldn't be passed to the AggregateRoot, the parameters should be extracted and passed into well defined methods.

Source/UserManagement/Domain/StaffUser/AggregateRoots/StaffUserManagement.cs
```diff
@@ -0,0 +1,161 @@
...
+                command.Email, command.YearOfBirth, (int)command.Sex, command.NationalSociety,
+                (int)command.PreferredLanguage, command.Location.Longitude,
+                command.Location.Latitude, command.Position, command.DutyStation
+                ));
+
+            if (command.MobilePhoneNumber != null && command.MobilePhoneNumber.Count > 0)
+            {
+                foreach (var number in command.MobilePhoneNumber)
+                {
+                    Apply(new PhoneNumberAddedToStaffUser(
```
smithmx: I don't have a problem with an Array on an event.

Source/UserManagement/Domain/StaffUser/CommandHandlers/IStaffUserCommandHandler.cs
```diff
@@ -0,0 +1,16 @@
+using doLittle.Runtime.Commands;
+using Domain.StaffUser.Commands;
+
+namespace Domain.StaffUser.CommandHandlers
+{
+    public interface IStaffUserCommandHandler : ICanHandleCommands
```
smithmx: No need for interface (see earlier comment).

Source/UserManagement/Domain/StaffUser/Commands/AddStaffUser.cs
```diff
@@ -6,23 +6,32 @@
...
+using doLittle.Commands;
 
-namespace Domain
+namespace Domain.StaffUser.Commands
 {
-    public class AddStaffUser
+    /// <summary>
+    /// Handles adding all kinds of staffusers.
+    /// It seems to that we have the option of this method, where fields can be empty.
+    /// Or have an add/remove command for each type of staffuser, which seems cumbersome and not dry.
```
smithmx: DRY never trumps your domain.  Anyway, DRY refers to logic, not data structures.  Don't be afraid to duplicate DTOs (which is what Commands and Events are).

If you want, you can create an abstract command and just inherit from it.  You can have a base validator that does all the work and can just be used by the concrete validators.

Source/UserManagement/Domain/StaffUser/Validators/AddPhoneNumberToStaffUserValidator.cs
```diff
@@ -0,0 +1,25 @@
...
+using Domain.StaffUser.Commands;
+using FluentValidation;
+
+namespace Domain.StaffUser.Validators
+{
+    public class AddPhoneNumberToStaffUserValidator : AbstractValidator<AddPhoneNumberToStaffUser>
+    {
+        public AddPhoneNumberToStaffUserValidator()
+        {
+            CascadeMode = CascadeMode.StopOnFirstFailure;
```
smithmx: Don't need a Global cascade mode.

Source/UserManagement/Domain/StaffUser/Validators/AddStaffUserValidator.cs
```diff
@@ -46,26 +37,47 @@ public AddStaffUserValidator()
...
                     //.Matches(@"^\(?([0-9]{3})\)?[-. ]?([0-9]{3})[-. ]?([0-9]{4})$")
                     .WithMessage("Mobile phone number is not correct - Has to be of correct form");
             });
 
+            When(_ => _.Role.RequiresAgeAndSex(), () =>
+            {
+                RuleFor(_ => _.YearOfBirth)
+                    .NotEmpty().WithMessage("YearOfBirth is not correct - Has to be specified");
+
+                RuleFor(_ => _.Sex)
```
smithmx: Be careful here.  An enum will have a default value if it is not passed (the MVC model builder will do this):

Source/UserManagement/Events/MessageGenerated.cs
```diff
@@ -6,7 +6,7 @@ namespace Events
 {
     public class MessageGenerated : IEvent
     {
-        public Guid Id { get; set; }
+        public Guid Id { get; set; } //QUESTION: einari, michael: What does this represent? An EventSourceId or the StaffUserId of the datacollector?
```
smithmx: It should be the EventSourceId  which is the Staff User Id.  There's no need to embed the id of the event in the body of the event.  Personally, I think Id is ambiguous and should be avoided.  StaffUserId is a better name.

Source/UserManagement/Policies/GreetingGenerators/GreetingGeneratorEventProcessor.cs
```diff
@@ -1,46 +1,45 @@
...
-            var smsGenerator = _greetingHistories.GetByPhoneNumber(@event.PhoneNumber);
+            var smsGenerator =await  _greetingHistories.GetByPhoneNumberAsync(@event.PhoneNumber);
             if (smsGenerator != null)
             {
                 return;
             }
 
-            var smsGeneratorAggregateRootRepository = _messageGeneratorsAggregateRootRepository.Get(@event.Id);
-            smsGeneratorAggregateRootRepository.GenerateMessage(new Domain.MessageGenerators.GenerateMessage()
+            var smsGeneratorAggregateRootRepository = _messageGeneratorsAggregateRootRepository.Get(@event.DataCollectorId);
```
smithmx: I think this is due to a limitation in the infrastructure, but I'd rather that the policy could be an Event Source and generate the event.  MessageGenerator as an AggregateRoot just feels wrong.

Source/UserManagement/Read/GreetingGenerators/GreetingHistoryEventProcessor.cs
```diff
@@ -11,17 +11,18 @@ public GreetingHistoryEventProcessor(IGreetingHistories greetingHistories)
         {
             _greetingHistories = greetingHistories;
         }
-
-        public void Process(PhoneNumberAddedToDataCollector @event)
+        
+        //TODO: QUESTION: Shouldn't this listen to MessageGenerated-event?
```
smithmx: I'm not sure I follow what's happening here.

If we want to generate a message when a phone number is added and then track the success of the message (know if it arrived), then I think the "MessageGenerated" event should contain more and should be the one that is tracked.

Source/UserManagement/Domain/StaffUser/Commands/AddStaffUser.cs
woksin: I see, that's smart!

Source/UserManagement/Domain/DataCollectors/Commands/AddDataCollector.cs
Velcrow81: Also, its called PhoneNumber in the events (and Origin by @einari in TextMessage and reports).

Source/UserManagement/Domain/StaffUser/AggregateRoots/StaffUserManagement.cs
```diff
@@ -0,0 +1,161 @@
...
+                        (int)command.Role
+                    ));
+                }
+            }
+            //TODO: DO the same for NationalSocieties
+        }
+        private void HandleAddDataOwner(AddStaffUser command)
+        {
+            Apply(new DataOwnerAdded(
+                command.StaffUserId, command.FullName, command.DisplayName,
```
Velcrow81: For code readability, you should add each parameter on its own line if you start breaking it up (personal preference, but also related to how the human eye moves when reading)

Source/UserManagement/Domain/StaffUser/AggregateRoots/StaffUserManagement.cs
Velcrow81: According to @einari it is better to have one event per phone number added (i.e. no arrays or as few as possible). That is at least my understanding from previous discussions

karolikl:
The failing builds are not your fault... I've added a fix for it here: #487

woksin:
@karolikl there are build issues as well in this PR, but Michael and I are working on it, making it pretty.

Source/UserManagement/Concepts/Language.cs
```diff
@@ -2,7 +2,7 @@ namespace Concepts
 {
     public enum Language
     {
-        English,
+        English = 0,
```
Davlind: Enums always initialize at 0, so not really necessary.

Source/UserManagement/Concepts/Role.cs
```diff
@@ -1,7 +1,7 @@
-﻿namespace Concepts{
+namespace Concepts{
     public enum Role
     {
-        Admin,
+        Admin = 0,
```
Davlind: Enums always initialize at 0, so not really necessary.

Source/UserManagement/Concepts/Sex.cs
```diff
@@ -2,7 +2,7 @@ namespace Concepts
 {
     public enum Sex
     {
-        Male,
+        Male = 0,
```
Davlind: Enums always initialize at 0, so not really necessary.

woksin:
This PR will be ready to be merged before the end of the day. We just need to do some cleanup first.

woksin:
This PR is ready for merging now.

Velcrow81:
Great work both of you 😊

karolikl:
Great job!!

Source/UserManagement/Domain/StaffUser/Registering/RegisteringCommandHandlers.cs
```diff
@@ -0,0 +1,70 @@
...
+{
+    public class RegisteringCommandHandlers : ICanHandleCommands
+    {
+        IAggregateRootRepositoryFor<StaffUser> _repository;
+        ISystemClock _systemClock;
+
+        public RegisteringCommandHandlers(IAggregateRootRepositoryFor<StaffUser> repository, ISystemClock systemClock)
+        {
+            _repository = repository;
+            _systemClock = systemClock;
```
Velcrow81: Why not datetimeoffset to handle timezones?

smithmx: Testability.  You can't verify the result of DataTimeOffset.UtcNow.  The implementation of SystemClock just delegates to that in anycase, but allows mocking.

PR #484, opened 2018-02-26
nginx proxy configuration

karolikl:
Changes done: 
- Added nginx ingress controller to route all traffic through the nginx proxy
- Added Helm chart so that all future deployments can be done with Helm
- In frontend Docker images, build using --base-href to serve static content from subpaths

After this PR has been completed, you can add 40.118.30.231 test.cbsrc.org to your hosts file and access all resources in the following way: 
test.cbsrc.org -> Admin frontend
test.cbsrc.org/users -> User Management frontend
test.cbsrc.org/reporting -> Volunteer Reporting frontend
test.cbsrc.org/adminbackend -> Admin backend
test.cbsrc.org/usermanagementbackend -> User Management backend
test.cbsrc.org/volunteerreportingbackend -> Volunteer Reporting backend

Next steps: 
- Add SSL certificate and create a record for test.cbsrc.org
- Deploy using Helm and Brigade
- Rename paths to e.g. /api/admin or similar when we have verified that the configuration works as expected, the current paths are rhubarb...
- Update list of maintainers, not quite sure who should be on it?
- Add all k8 yaml files to the Helm chart templates folder.

### DIFF
--- a/Deploy/templates/admin.yaml
+++ b/Deploy/templates/admin.yaml
@@ -41,7 +41,7 @@ spec:
     spec:
       containers:
       - name: cbs-admin-backend
-        image: karolikl/cbs-admin-backend:{{services.admin.backend.version}}
+        image: karolikl/cbs-admin-backend:{{ .Values.services.admin.backend.version }}
         ports:
         - containerPort: 80
 ---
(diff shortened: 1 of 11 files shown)


Deploy/templates/admin.yaml
```diff
@@ -0,0 +1,86 @@
...
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: cbs-admin-frontend
+        environment: dev
+    spec:
+      containers:
+      - name: cbs-admin-frontend
+        image: karolikl/cbs-admin-frontend:{{ .Values.services.admin.frontend.version }}
```
Velcrow81: Is his in your container registry?

karolikl: Yes, we started moving it to another image repository during the last codeathon, but we didn't have the correct access rights. Will make sure we update asap :)

PR #534, opened 2018-03-09
Minor refactor to improve readability in text message processor

### LINKED ISSUE #533: Refactor Text message processing to more readable
It needs to be clearer what hapens if data collector is null. Now it seems like it will fail

Velcrow81:
Fixes #533


Velcrow81:
Seems to be a problem with nuget restore, but I can't see that I have changed anything related to this. Anyone have an idea? 

Cake.exe : An error occurred when executing task 'Restore-NuGet-Packages'.
459At C:\projects\cbs-g81xy\Build\build.ps1:233 char:1
460+ &$CAKE_EXE $cakeArguments
461+ ~~~~~~~~~~~~~~~~~~~~~~~~~
462    + CategoryInfo          : NotSpecified: (An error occurr...uGet-Packages'.:String) [], RemoteException
463    + FullyQualifiedErrorId : NativeCommandError
464 
465Error: 
466One or more errors occurred.
467
468
469	
470.NET Core CLI: Process returned an error (exit code 1).
471
472
473Command executed with exception: 
474

karolikl:
Looks like the package dolittle.events.processing cannot be found: 
"C:\projects\cbs-g81xy\Source\VolunteerReporting\Web\Web.csproj : error NU1101: Unable to find package doLittle.Events.Processing. No packages exist with this id in source(s): aspnetcore-ci-dev, dolittle-ci, nuget.org "

I'm not able to find it when I search nuget.org, is that where the package should be located, @einari?

karolikl:
Update, the package used to be on MyGet: https://www.myget.org/gallery/dolittle 
Looking at a cached view of the feed, the doLittle.Events.Processing package used to exist there: 
https://webcache.googleusercontent.com/search?q=cache:BSXzQ-vnuX4J:https://www.myget.org/gallery/dolittle+&cd=2&hl=en&ct=clnk&gl=no 

But in the current feed, it's gone. I see most of the dolittle packages on MyGet were updated 1 day ago, so I suspect this is when it disappeared...

I've created an issue on this: #535

Velcrow81:
Thanks @karolikl 😊

karolikl:
Our build issues are now fixed, could you please push a new dummy commit to trigger the PR builds again?

Source/VolunteerReporting/TextMessaging/TextMessageProcessor.cs
```diff
@@ -48,131 +47,83 @@ public class TextMessageProcessor : ICanProcessTextMessage
...
+                    message.Sent);                
+                return;
             }
-            else
+
+            var healthRiskReadableId = parsingResult.HealthRiskReadableId;
+            var healthRiskId = _healthRisks.GetIdFromReadableId(healthRiskReadableId);
+            if (healthRiskId == HealthRiskId.NotSet)
+            {
+                var errorMessages = new List<string> { $"Unable to find health risk, since there are no health risks with a readable id of {healthRiskReadableId}" };
```
woksin: errorMessages is not used, shouldent this list be appended to parsingResult.ErrorMessages and passed as argument to caseReporting.Report* ?

Source/VolunteerReporting/TextMessaging/TextMessageProcessor.cs
```diff
@@ -48,131 +47,83 @@ public class TextMessageProcessor : ICanProcessTextMessage
         public void Process(TextMessage message)
         {
             var parsingResult = _textMessageParser.Parse(message);
+            var isTextMessageFormatValid = parsingResult.IsValid;
             var dataCollector = _dataCollectors.GetByPhoneNumber(message.OriginNumber);
+            var unknownDataCollector = dataCollector == null;
             var caseReportId = Guid.NewGuid();
             var caseReporting = _caseReportingRepository.Get(caseReportId);
```
woksin: This class looks fine, but two minor things that botter me:
1) I would prefer to have private methods in the class to handle the different cases, instead of having to use the return statements. (Minor thing really)
2) Shouldn't we here instead of calling on the AggregateRoot for CaseReport emit a command to a CaseReport command handler? 

Other than that, it look fine. An improvement for sure.

Source/VolunteerReporting/TextMessaging/TextMessageProcessor.cs
Velcrow81: Ups :) Nice catch! Copy paste error

Source/VolunteerReporting/TextMessaging/TextMessageProcessor.cs
Velcrow81: 1. Interesting thought. I tried to do something similar, but ended up with this since it is easy enough to follow. How would you write the private methods that would ensure that only one case report was emited?
2. We have gone over this before and we are still not quite sure how this should be. The reason that we do it this way, is that sending an SMS can be seen as the command. If we recieve the SMS, that is a fact, something that happend. As you can see, this logic happens in a Process method, which means that we cannot recject it based on i.e. input validation. If the SMS is wrong, we must handle it by emiting the corresponding event. Make sense?

woksin: 1) I have no concrete suggestion at the moment, I just stop and think when I see multiple return statements in a method.

2) I see your point, Bjørn. It may be beneficial to look at textmessage processing as it's own thing entirely even.

PR #537, opened 2018-03-11
Added linter (tslint) for Volunteer Reporting web application

beatfactor:
This PR adds support for source code linting with [tslint](https://github.com/palantir/tslint) to enforce a consistent code style for the angular frontend code. The existing rules are inspired from angular's own code style.

Also fixed existing code style issues.

--
Main rules:
- _Indentation_: 2 spaces
- _Quote mark:_ single quotes, except to avoid escaping
- _Semicolons_: always use

**Important:**
Docker will refuse to build if the code doesn't satisfy the lint requirements (i.e. if `ng lint` fails).

Example failure output:
```
---> 9b25098724a4
Step 9/12 : RUN ["yarn", "lint"]
 ---> Running in 26b378956e96
yarn run v1.5.1
$ ng lint

ERROR: /src/src/app/app.module.ts[10, 26]: " should be '

Lint errors found in the listed files.
error An unexpected error occurred: "Command failed.
Exit code: 2
Command: sh
Arguments: -c ng lint
```


alexanderhoset:
May be an idea to move to https://standardjs.com/  to make lint rules for all contexts easy @beatfactor ?

beatfactor:
Maybe, but I think this is enough for what we need. I haven't used it before, because I believe that adding an extra tool makes it less transparent what the linter does and it's not possible to configure, in case we want to deviate from their coding style. Which btw, it promotes that we never use semicolons at the end of the statement, and I find that a little weird.

alexanderhoset:
Agree, just wanted your opinion on it

karolikl:
@beatfactor: Our build issues are now fixed, could you please push a new dummy commit to trigger the PR builds again? Sorry about the inconvenience!

PR #541, opened 2018-03-11
Remove Linux Docker image.

sheeeng:
Clean up the Linux image, which is not taken into use.


sheeeng:
I see that CBS enable TravisCI already.
I will remove `.travis.yml` file since it is not used at the moment.
Should we continue to support the Linux image?

Velcrow81:
Please commit a dummy request or merge in latest to trigger new build. There was a build error before that has been fixed

sheeeng:
I got the impression earlier that we are not using TravisCI anymore, but focusing only on using AppVeyor?

Velcrow81:
I am not familiar with this. I think we need both to cover Linux and Windows. Any input @karolikl ?

einari:
Kubernetes is running on Linux and all containers are based on that - so it would be nice to have it confirmed running on Linux. AppVeyor is bringing out Linux support however - so maybe we could migrate over to it? Read more [here](https://www.appveyor.com/beta/).

sheeeng:
Thanks for the clarification.

Issue #500, opened 2018-03-05
Update Web/TestGenerator to reflect changes to the StaffUser system


woksin:
The TestGenerator should create commands for adding all the different kinds of Staffusers

Issue #479, opened 2018-02-23
Create a TestDataGenerator mechanism for the Web.Api


woksin:
Make something similar to what we have in VolunteerReporting so the project's functionality can be tested. 

Very useful for testing the backend and frontend of the project.

woksin:
#482 Solves most of the use cases currently for the TestDataGenerator

PR #549, opened 2018-03-17
Make Testdata generator simpler and more flexible

woksin:
Should close down issues: #479, #500 

Made the Web project dependent on Domain.Specs, the though is that the testdata generator should have access to the specs of the domain to be able to always generate correct data according to the specs.

Also, should methods in the Read collections really throw exception, for example when a user with a specific ID is not found?

### DIFF
--- a/Source/UserManagement/Domain.Specs/StaffUser/Registering/given/commands.cs
+++ b/Source/UserManagement/Domain.Specs/StaffUser/Registering/given/commands.cs
@@ -1,6 +1,5 @@
 using Domain.StaffUser;
 using Domain.StaffUser.Registering;
-using System.Collections.Generic;
 using System;
 using Concepts;
 using doLittle.Commands;
--- a/Source/UserManagement/Read/DataCollectors/DataCollectorEventProcessor.cs
+++ b/Source/UserManagement/Read/DataCollectors/DataCollectorEventProcessor.cs
@@ -34,6 +34,7 @@ namespace Read.DataCollectors
 
         public void Process(DataCollectorUpdated @event)
         {
+            //TODO: Should be replaced with a new system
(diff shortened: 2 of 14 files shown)


einari:
The Web project shouldn't be dependant on the Specs project - but rather if there is something that we want to share between them, put it in a third one. 
 
As for exceptions - in general a rule of thumb; exceptions are to be considered an exceptional non-recoverable state and something we should avoid being in. If a query can't find any results, it shouldn't throw an exception - but rather an empty collection.

woksin:
@einari I see. I'll change that then, I was wondering if this was okay so thanks for the feedback.

And for the exceptions, so returning null instead is preferred over exceptions?

einari:
Great. 

Returning null is probably not a good idea - null is evil :)

But if it's supposed to return a collection, return an empty one.

If it is returning a single instance and it is wrong there is something missing to validate before you get to the place where you're supposed to return this. 
We (Dolittle) have specific return types for the different pipelines.

einari:
https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare

woksin:
I've read it :wink: I've got mixed feelings, but what do I know...

einari:
What is missing is a semantically correct response. For instance what we return is a CommandResult and a QueryResult object for the different pipelines and it contains broken rules avoiding it all together. Sometimes there are exceptions - but then theee is a good reason

woksin:
I agree with you there, how we're actually going to implement this is a theme for discussion for code eve I think

woksin:
I'm thinking of removing the "test stuff" from the Web-project entirely and putting it in its own project

Issue #496, opened 2018-03-04
Read Model Projections


smithmx:
From the new events after refactoring of StaffUsers, we need to reimplement the Read Models and projections from the Event Processors.

woksin:
#530 
Functionality left to be implemented:
  Adding phonenumbers
  Adding national societies
  Controller endpoints

PR #581, opened 2018-03-26
#496 Implemented basic query system for read models in User Management

woksin:
I have implemented, and tested to some extent, queries for the read models in the UserManagement project. 

Created endpoints for these queries, though these will eventually be deprecated when we can use the nice voodoo magic that we're going to get soon. (Query and Command endpoints, @einari :-) ) 

I'm unsure on the QueryProviderFor<..> classes, I get that they should be QueryProviders for something that is "Queryable", but I tried alot of things, but since we are working Mongo Collections,  I could not figure out how to do it in a nice way... Some input, @einari ? 

Also, async queries, @einari, should that be a thing? Does the system make it possible for async queries, and is it wanted?


einari:
w0000t..    Impressive work @woksin..  
Question, does the `IAsyncCurstor<>` interface have a non-generic version that it implements as well? If so, one could create a generic implementation of this provider that could support all types. Much like you find for the `IQueryable<>` implementation [here](https://github.com/dolittle/Bifrost/blob/master/Source/Bifrost/Read/QueryableProvider.cs).

If so, we should move the implementation into infrastructure and keep it there so that everyone automatically gets the benefit of it. 

Also, probably starting today with transitioning to the latest and greatest of Dolittle, which has the official endpoints for Commands and Queries - and in a few days we'll have JavaScript wrappers for these as well in the form of two classes; CommandCoordinator and QueryCoordinator for the frontend.

Source/UserManagement/Read/DataCollectors/DataCollector.cs
```diff
@@ -10,7 +10,7 @@ namespace Read.DataCollectors
     public class DataCollector : IReadModel
     { 
         [BsonId]
```
einari: Not part of this commit - but a comment in general. Mapping should not be done in the actual object, but use the Mongo mapping classes:
https://mongodb.github.io/mongo-csharp-driver/2.0/reference/bson/mapping/

Source/UserManagement/Read/DataCollectors/DataCollector.cs
```diff
@@ -10,7 +10,7 @@ namespace Read.DataCollectors
...
-        public Guid Id { get; set; }
+        public Guid DataCollectorId { get; set; }
```
einari: DataCollector is an important concept in the domain, should introduce a ConceptAs<> for it in a Concepts project.

Source/UserManagement/Read/StaffUsers/Models/BaseUser.cs
```diff
@@ -10,14 +10,13 @@ namespace Read.StaffUsers.Models
     public abstract class BaseUser : IReadModel
     {
         [BsonId]
-        public Guid StaffUserId { get; }
+        public Guid StaffUserId { get; set; }
         [BsonRequired]
```
einari: General comment, again; mapping -> separate concern, separate classes:
https://mongodb.github.io/mongo-csharp-driver/2.0/reference/bson/mapping/

Source/UserManagement/Read/StaffUsers/UserEventProcessor.cs
```diff
@@ -102,10 +102,6 @@ public UserEventProcessor(IStaffUsers collection)
         public void Process(PhoneNumberRegistered @event)
         {
             var baseUser = _collection.GetById<BaseUser>(@event.StaffUserId);
-            if (baseUser == null)
```
einari: Excellent!  Great that you remove these - we have to assume that validation and/or business rules on the Command side was done correctly and that this code can naively carry on.

Source/UserManagement/Read/DataCollectors/DataCollector.cs
woksin: I see, single responsibility principle in its fullest I guess :wink: I'll look into it

Source/UserManagement/Read/DataCollectors/DataCollector.cs
woksin: On it

Source/UserManagement/Read/DataCollectors/DataCollector.cs
einari: 👍  Its actually separation of concerns; one being the representation of the readmodel, while the other being a very technology specific mapping and description needed as requirement for Mongo.

You see the similar thing with validation in vanilla ASP.NET MVC where attributes are being used. Validation being a separate concern all together.

When one separates the concerns, it is easier to identity patterns that could then result in the concept of cross-cutting concerns. Patterns are then mapped into conventions and one can have rules in the codebase that things will automatically pick up on instead. Its also possible to apply things cross-cuttingly in a specific manner as well. For instance if you're using a Concept type representing something, you could have an Input validation rule associated with it and it could then be automatically discovered and hooked up whenever it is in use for a Command for instance.

woksin:
@einari The problem I seemed to have was that the collection the IAsyncCursor pointed to gets disposed before it's given to the query provider, I don't know how to work around that.

Issue #595, opened 2018-04-04
Show phone number for unknown data collectors


karolikl:
If a case report is received from an unknown number, we need to display the phone number (in the data collector column). 

![image](https://user-images.githubusercontent.com/5235776/38332052-4dae1310-3855-11e8-9d92-a0a279a218a0.png)

woksin:
I can fix this 👍

woksin:
Should work because of #598 .
Can be closed when we can confirm this is working in the production environment.

Issue #596, opened 2018-04-04
Display detailed error message on case reports


karolikl:
When there is an error in parsing case reports, we need to somehow display the detailed error message: 

![image](https://user-images.githubusercontent.com/5235776/38332178-a38a203a-3855-11e8-9a4f-9b24cd7b01c6.png)

woksin:
I can fix this 👍

woksin:
Should work because of #598 .
Can be closed when we can confirm this is working in the production environment.

PR #598, opened 2018-04-06
Should solve #595 and #596

woksin:
This PR should resolve issues #595 and #596.

Had to do some work arounds in the Read, callin Synchronous methods instead of Async.

Important! When writing read models, mongodb will have issues when serializing/deserializing _id fields that are classes in the models (for example if we use a concept for the field in a model which represents its id field into the DB). Use Guid, that's atleast my temporal solution.

Had a pretty bad time solving this because of the thing mentioned above, so I changed some signatures in some of the Read-collections to be synchronous instead of Async. I'll just leave it as it is.

Another thing to mention is that when I had two EventProcessor classes listening for the same event, with async Task Process methods the results could vary, meaning that sometimes (or all the time in my case) only one of the EventProcessors catched the event... @einari

### DIFF
--- a/Source/VolunteerReporting/Web.Angular/src/environments/environment.ts
+++ b/Source/VolunteerReporting/Web.Angular/src/environments/environment.ts
@@ -5,5 +5,5 @@
 
 export const environment = {
   production: false,
-  api: 'http://dev.cbsrc.org/volunteerreportingbackend'
+  api: 'http://localhost:5000'
 };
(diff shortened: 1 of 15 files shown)


Source/VolunteerReporting/Web.Angular/src/environments/environment.ts
```diff
@@ -5,5 +5,5 @@
 
 export const environment = {
   production: false,
-  api: 'http://dev.cbsrc.org/volunteerreportingbackend'
+  api: 'http://localhost:5000'
 };
```
woksin: It needs to be  localhost;5000, right @karolikl ?

einari:
The event processors are being governed thread wise without really considering whether or not the method is async. We've traditionally not made them async because of this. So there might be some need to investigate this further. Will make an issue on it.

As for concepts and serializing and Mongo - we have an implementation of that here:
https://github.com/dolittle-extensions/ReadModels.MongoDB/tree/master/Source

Source/VolunteerReporting/Web.Angular/src/environments/environment.ts
karolikl: Correct :)

PR #613, opened 2018-04-10
Cleaning up deployment files

karolikl:
All deployment files have been moved to the /Deploy folder, so I'm removing the project specific deploy folders to avoid any future confusion.


jakhog:
The image-tags should be updated again... If you remove the -... part, I think it should be future proof :)

PR #615, opened 2018-04-10
Dropping iFrame for navigation for now - composing in as component 

### LINKED ISSUE #373: Set up a new web frontend for navigation bar
### Description
We need a navigation bar for navigating across the different solutions. This will also wrap them in so they will be precept as one solution.

### LINKED ISSUE #402: Make navigation project deployable
### Description
We need this project to be deployable so that it can be used.

The deployable parts will be the output in ./dist after `webpack -p`

### LINKED ISSUE #516: Add navigation bar to User Managment
When the nav bar is done in Admin in #515 , this can be replicated in User Managment

### LINKED ISSUE #517: Add navigation bar to Volunteer Reporting
When the nav bar is done in Admin in #515 , this can be replicated in User Managment

einari:
Closes #517 
Closes #516
Closes #373 
Closes #402 

Still some work left related to this - will put in a frontend thing for now for #601 for instance, 
plus recreate the logic I had done in Knockout earlier in Angular.


einari:
The reason why it fails on AppVeyor is that it can't find the navigation bar component - which is sitting behind a symbolic link. 

`[00:04:11] ERROR in Error: Error encountered resolving symbol values statically. Could not resolve ./navigation/nav-top-bar/nav-top-bar.component relative to `

Since I'm don't have Windows here - I can't really reproduce it - but I'll add something to the build script that will hopefully fix this, based on the documentation in ["git-for-windows" wiki](https://github.com/git-for-windows/git/wiki/Symbolic-Links).

PR #625, opened 2018-04-11
Cleaning up sub-navigation of admin

karolikl:
Cleaning up sub-navigation of admin to make it look a bit nicer before the field test on Friday. I've also added Healthrisks to the navigation, and set the project list as default route.


karolikl:
Sorry, it's getting late.... I branched out from #624, so please have a look at that PR before this one as this contains the same changes.

Issue #656, opened 2018-04-17
Populate Message property when SMS is correctly parsed


karolikl:
Today, the Message property is null when an SMS is correctly parsed, while it is populated with the SMS content when it is incorrectly parsed. 

We need Message to be populated for correctly parsed case reports as well.

woksin:
@karolikl This field has been correctly populated in the past, does this now happen every time we send a correctly formated sms?

karolikl:
Yes, it happens every time we send a correctly formatted SMS, Message is always null.

woksin:
My gut feeling tells me that #657 and #656 are related. I can take a look at this tomorow, unless anyone else wants to

woksin:
Should be fixed with #662,

PR #662, opened 2018-04-18
Fix for #656

woksin:
Fixes the issue and also fixes an error where the test data DataCollectors doesn't get serialized properly into DataCollectorRegistered Events.


jakhog:
The instructions on how to add changes are at the bottom of the comments here if you don't know how (I didn't).
`Add more commits by pushing to the #656 branch on woksin/cbs.`

PR #663, opened 2018-04-18
Text message parsing cleanup

woksin:
Made the parsing of text messages more explicit and describing, in my opinion.


Source/VolunteerReporting/Concepts/AgeGroup.cs
```diff
@@ -0,0 +1,8 @@
+namespace Concepts
+{
+    public enum AgeGroup
+    {
+        AgeUnder5 = 1,
+        AgeOver5 = 2
```
Velcrow81: Should be Aged5AndOver for 2 according to latest naming of reports

Issue #359, opened 2018-01-06
Export Case Reports 


alexanderhoset:
Ref cbs/Design/cbs.pdf (" Case Reports export")
  
- Edit (woksin) added task list
- [x] Export to csv
- [x] Export to excel
- [x] Export to pdf

woksin:
Do we want to combine all Case Reports into one table, or give the user the option to choose which subset of the case reports he/she wants?

alexanderhoset:
subset, as in if you have applied a filter, that has to count 
so export the current list  with filter or not. Normally they will use the daterange filter before selecting export

woksin:
@alexanderhoset On the coreteam sync you said that you would look into this export functionality, Should we assign this to you?

jakhog:
What's the plan here, server-side or client-side?

woksin:
@jakhog I don't know what the others are thinking, but personally I think that this should be a server-side job

eprom:
I go with server side.

aninek:
We need to include some geographical information beyond GPS lat-long in the MVP as there is no analytics/dashboard. Therefore, when data collector registration is updated, can we include in the export also 

- Region

- District

- Village
Linked to #695

aninek:
Timestamp needs to be split in date and time also in export.

karolikl:
Fixed by #687 and #695

PR #672, opened 2018-04-23
#359 Export to pdf

woksin:
Export to pdf backend functionality should be in place.

Location of datacollector should now appear even if it's an invalid report.

Be aware, testing the export functions in swagger results in (at least for me) corrupted files for some reason.
Do the request by manually writing the route.

I only did local testing of this using localhost:5000/api/casereports/export/pdf and csv


Source/VolunteerReporting/Web/Web.csproj
```diff
@@ -10,6 +10,7 @@
   <ItemGroup>
     <PackageReference Include="DocumentFormat.OpenXml" Version="2.8.1" />
     <PackageReference Include="FakeItEasy" Version="4.1.0" />
+    <PackageReference Include="iTextSharp" Version="5.5.13" />
```
woksin: I think the license for ITextSharp is fine, someone with more insight into that kind of stuff needs to look at that.

woksin:
No one merge yet, there is potentially breaking stuff in here since I added a property to an event, need to hear with @einari

Velcrow81:
Adding properties to events should be safe. Just remember to handle existing read models that dont have the value by adding default vales or make them optional. Correct med when I am wrong @einari 😊. And I am sure dolittle have some fancy way of fixing this through versions etc

einari:
Adding properties are "safe". 
Right now when we have not put it into production - anything goes. 
When we are in production - we need to be a bit more aware. If it's adding a new property and it is not something another BC relies on for meaning till it gets the property itself and is up to speed - it's OK.

We have a generation system that handles generation transitions of events. This needs a bit of attention with regards to inter BC connections.

einari:
What should be a governing thing though; adding properties to events might not necessarily be what you're looking for. Events are not vessels for read models, with added requirements - one should keep an eye on whether or not it is a new event rather than introducing a property as it might be new domain knowledge.

PR #674, opened 2018-04-24
#359 Export to Csv + Export button

woksin:
Implemented the functionality for exporting case reports to csv in the backend.

Created the export component that can export all case reports.

Coulnd't quite figure out how to download the file sent from the asp.net controller in angular in a good way, for now I just call window.open on the url ( this is perhaps very naive ).

Export with filtering and ordering (again, quite naive in my opinion. And it will probably change when @einari decides to drop the DoLittle2.0-bomb)

### DIFF
--- a/Source/VolunteerReporting/Web.Angular/package-lock.json
+++ b/Source/VolunteerReporting/Web.Angular/package-lock.json
@@ -9993,6 +9993,11 @@
         }
       }
     },
+    "yarn": {
+      "version": "1.6.0",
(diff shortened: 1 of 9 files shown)


woksin:
I would like to know what caused the failed check, @jakhog @einari ?

jakhog:
Are you not able to see the test details @woksin? I thought they were public. Anyways, the errors are linter based. You should see them if you run `yarn lint`, I think. For shure you will see them if you build the Docker image...

On another note, I have also been working on the frontend, but quite differently from yours. Maybe we can consolidate a bit before this is merged? Mostly details, but we can have a chat on slack perhaps?

PR #675, opened 2018-04-24
Update VR case report export frontend and fixing bugs in backend

jakhog:
- Merged the two different frontend implementations
- Added selection of fields to export in frontend (needs backend implementation)
- Refactored backend export classes to share more of the functionality
- Fixed some small bugs in backend exporters


woksin:
Nice refactoring. Did you fix the issues with csv and pdf?

PR #709, opened 2018-05-08
Introducing new standard interface repositories + Export data collectors

woksin:
- The entities in the database collections are now IReadModel

- The Repositories are IReadModelRepositoryFor<T>

- Implemented a standard implementation of a repository that all IReadModelRepositoryFor can inherit from so that they all share a common set of operation and so that it's more user friendly. 

- Introduced IMongoDbClassMap<> and an abstract implementation MongoDbClassMap<> in Infrastructure.Read. All ReadModels in a solution's Read namespace should specify a ClassMap-class that extends MongoDbClassMap<> for automatic BsonClassMapRegistration through IReadModule.

- Introduced IReadModule and implementation ReadModule in Infrastructure.Read for automatically map BsonClassMaps for IReadModels that has a MongoDbClassMap-class defined in Read.

- Corrected Queries for the ReadModels in UserManagement

- Created export functionality for DataCollectors in Usermanagement frontend.

I've done some testing, and it seems like the Read-side is working as usual for all the solutions


Source/VolunteerReporting/Read/CaseReportsForListing/CaseReportForListing.cs
```diff
@@ -2,39 +2,39 @@
...
+        public int NumberOfFemalesAged5AndOlder { get; set; }
+        public DateTimeOffset Timestamp { get; set; }
+        public Location Location { get; set; }
+        public string Origin { get; set; }
+        public IEnumerable<string> ParsingErrorMessage { get; set; }
+
         //TODO: Remove properties below from model once data is migrated to new property names or can be removed
         //This feels dirty. Waiting for input from @einari on how to handle backwardscompability better
         [Obsolete]
-        public int? NumberOfFemalesAgedOver4 { get; internal set; }
```
einari: If these are [obselete] we can take them out. We don't have a previous version of the software running.

woksin:
Thanks for the thorough review, @einari.

Putting the generic Repository in it's own infrastructure project is a good idea, I think that I will do that.

Do you mean that we should only have Async methods? Is that safe? I've experienced trouble with using async methods with mongo in this project before.

woksin:
I could remove the whole generic ID concept and be cheeky and just expose get with filter and expression predicate (?)

einari:
Great. 
The async problem we've seen could that be related to Dolittle assuming synchronosity in some areas? Dolittle is doing a bad assumption on this, one particular place would be the event processors for this. I've registered an issue with Dolittle for this: https://github.com/dolittle/DotNET.SDK/issues/20

Having a filter as you say with expression predicate would be great.

Source/Infrastructure/Read/ExtendedReadModelRepositoryFor.cs
```diff
@@ -0,0 +1,427 @@
...
+            {
+                var propInfo = GetIdProperty(entity);
+                object id = propInfo.GetValue(entity);
+
+                return GetIdAsBsonValue(id);
+            }
+            catch (InvalidOperationException)
+            {
+                var bsonDocument = entity.ToBsonDocument();
+                return bsonDocument["_id"];
```
woksin: If ReadModel has property Id, that is automatically seen as _id. If not, convert it to BsonDocument and retrieve _id (assumes that we have provided a BsonClassMapping for the id for the ReadModel)

Comments, @einari ?

Source/Infrastructure/Read/IExtendedReadModelRepositoryFor.cs
```diff
@@ -0,0 +1,132 @@
...
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using System.Threading.Tasks;
+using Dolittle.ReadModels;
+using MongoDB.Driver;
+
+namespace Infrastructure.Read
+{
+    public interface IExtendedReadModelRepositoryFor<T> : IReadModelRepositoryFor<T>
+        where T : IReadModel
```
woksin: A big interface, but it provides most of the same methods that IMongoCollection provides.

Source/VolunteerReporting/Web/Startup.cs
```diff
@@ -15,6 +16,12 @@ public class Startup : Infrastructure.AspNet.Startup
             IHostingEnvironment env,
             IConfiguration configuration) : base(loggerFactory, env, configuration)
         {
+            RegisterReadModelBsonClassMaps();
```
woksin: Registering of the ReadModel BsonClassMaps

woksin: This is changed to register a IReadModule service instead. See newer comments

Source/Infrastructure/Read/IReadModule.cs
```diff
@@ -0,0 +1,15 @@
...
+    /// <summary>
+    /// In startup, create an override of public void ConfigureServicesCustom(IServiceCollection services).
+    /// Register IReadModule with s => new ReadModule(AppDomain.CurrentDomain).
+    /// Then instantiate the service with services.BuildServiceProvider().GetService<IReadModule></IReadModule>
+    ///     ();
+    /// </summary>
+    public interface IReadModule
+    {
+        void RegisterReadModelClassMaps(AppDomain appDomain);
+    }
```
woksin: In Startup.Cs in each solution's Web project we instantiate this with, comments @einari ?

public override void ConfigureServicesCustom(IServiceCollection services)
        {
            services.AddSingleton<IReadModule>(s => new ReadModule(AppDomain.CurrentDomain));
            services.BuildServiceProvider().GetService<IReadModule>();
        }

Source/Infrastructure/Read/MongoDbClassMap.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    public abstract class MongoDbClassMap<T> : IMongoDbClassMap<T>
+        where T : IReadModel
+    {
+        protected MongoDbClassMap()
+        {
+            if (!BsonClassMap.IsClassMapRegistered(typeof(T)))
+                BsonClassMap.RegisterClassMap<T>(Map);
+        }
```
woksin: All ReadModels in Read also provide a ClassMap- class that extends MongoDbClassMap.
And when IReadModule is instantiated in Startup, all the ClassMaps are automatically registered on construction

Source/Infrastructure/Read/ReadModule.cs
```diff
@@ -0,0 +1,23 @@
...
+            RegisterReadModelClassMaps(appDomain);
+        }
+
+        public void RegisterReadModelClassMaps(AppDomain appDomain)
+        {
+            var types = appDomain.GetAssemblies().FirstOrDefault(a => a.FullName.Split(',').First() == "Read")
+                        .GetTypes();
+            types.Where(t => t.IsClass && !t.IsAbstract && t.HasInterface(typeof(IMongoDbClassMap<>)))
+                .ForEach(classMapType => Activator.CreateInstance(classMapType));
+        }
```
woksin: Does discovery on the AppDomain for namespace Read. Creates an instance of all classes of IMongoDbClassMap in Read

Source/Infrastructure/AspNet/Startup.cs
```diff
@@ -46,8 +46,8 @@ public void ConfigureServices(IServiceCollection services)
                 .AddMvc();;
             services.Configure<ConnectionStringsOptions>(_configuration);
 
-            _bootResult = services.AddDolittle(_loggerFactory);
-
+            //_bootResult = services.AddDolittle(_loggerFactory); // Sometimes it seems like this goes in an infinite loop with the logger factory for some reason
+            _bootResult = services.AddDolittle(); // this works
```
woksin: For some reason, sometimes when I try to run a debug session of UserManagement or Admin, this goes in an ininite loop seemingly because of the _loogerFactory. When I remove the argument from the call it works without problem.
@einari

einari:
This pull request is WAY too big. I'm having a hard time getting through it all - lets not have this in the future. 

A lot of great work here - but there is a couple of themes I'd like to address:

`IInstancesOf<>` and `IImplementationsOf<>` are two interfaces that enables an easy way to discover instances or types. They both use the underlying `ITypeFinder` and also for making instances it uses the IoC container. You can find an example of usage [here](https://github.com/dolittle/DotNET.Fundamentals/blob/master/Source/Mapping/Maps.cs).

From an infrastructural perspective, we're trying to make the developers write as little plumbing as possible. This means that anything that can be automated that are repetitive tasks should be done by the infrastructure, not forcing the developer to do this. This means that we have to come up with supported conventions that can - if needed - be overridden by concrete implementations. For instance for the Bson class maps - all of the read models thus far does not really need an explicit class map definition. This basically because everyone has `AutoMap()` and an `IdMember`which is pointing to a member of the same name; `Id`. A very great scenario for automation and convention. In fact, as I mentioned as a comment, the MongoDB driver has support for conventions like this - and you don't need to have to implement things. 

Another theme I'm seeing; we don't need to anticipate things - only build whats really needed when it is needed. Don't anticipate infrastructure. The read model repository is in my opinion way too big and should be kept very simple. It does not need to abstract away everything that MongoDB supports. When we start using `IQueryFor<>`more actively - the queries should provide the functionality needed to perform the query. Meaning that all the generic repository things probably are not needed at all - we just need a point of querying. This is also true for when we start using the read models directly - all we want is to get single read models based on matching a specific property - often the `Id`. 

Lets get these things sorted so we can get this pull request in :).  
Love for it to not grow anymore.

woksin:
@einari I took your comments to heart, and I think that my new implementations is closer to what you suggested.
Take a look at these classes to get an overview of the changes:

https://github.com/woksin/cbs/blob/generic-repositories/Source/Infrastructure/Read/IMongoDbClassMapFor.cs

https://github.com/woksin/cbs/blob/generic-repositories/Source/Infrastructure/Read/IReadModule.cs

https://github.com/woksin/cbs/blob/generic-repositories/Source/Infrastructure/Read/ReadModule.cs

https://github.com/woksin/cbs/blob/generic-repositories/Source/Infrastructure/AspNet/ContainerBuilderExtensions.cs

https://github.com/woksin/cbs/blob/generic-repositories/Source/Infrastructure/AspNet/ServiceCollectionExtensions.cs

https://github.com/woksin/cbs/blob/generic-repositories/Source/Infrastructure/AspNet/Startup.cs

woksin:
When I try to build usermgmt frontend with docker locally I get an error when running the ng build --prod command. Seems like ugilyfyjs breaks?

I don't think that there's a problem with this PR specifically, I even tried to build the docker image in my local Master branch (which is sync with this Master-branch) and I get the same error.

@jakhog @einari @eprom

woksin:
I tried to replace RUN ["yarn", "install"] with RUN ["npm", "install"] in UserManagement/Web.Angular/Dockerfile
That seemed to work, the build succeeded. Maybe yarn install does something weird?

PR #721, opened 2018-06-05
Fixes an error where I could not build docker images. using npm insta…

woksin:
…ll instead of yarn install seems to fix the problem, but the build time is increased
This should also fix the problem with my other Pull Request not building the usermgt-frontend in docker


woksin:
Hmm, looks like this is also failing the UserManagement-frontend build.
Running docker build -f Source/UserManagement/Web.Angular/Dockerfile .
locally works. 

I cannot access the docker build logs or details, I guess that it would be nice if we could have a look at what's happening in the docker build.
@einari @karolikl @jakhog

woksin:
Close this PR, it seems like the issue fixed itself...

PR #722, opened 2018-06-06
MongoDBRegistrationSource now uses Dolittle.ReadModels.MongoDB's Conn…

woksin:
…ection and Configuration classes to setup the MongoDB connection. Provides automatic registration of ConceptSerializer


woksin:
We don't want this

PR #727, opened 2018-06-11
Split time

### LINKED ISSUE #687: Timestamp split in two fields
We need time stamp to be split into 2 separate fields: DATE and TIME. 
This will enable easy manual analysis and reporting before the dashboard/reporting features are added.

woksin:
Closes #687

### DIFF
--- a/Source/Infrastructure/AspNet/Initialization.cs
+++ b/Source/Infrastructure/AspNet/Initialization.cs
@@ -30,16 +30,6 @@ namespace Infrastructure.AspNet
                 .Enrich.FromLogContext()
                 .WriteTo.JsonConsole();
 
-            //TODO: This seemed to fix the problem where Amin was the only BC that would not startup because of an infinite logging loop.
-            // This problem indicates to me that we have to take a look at the logging infrastructure
-            if (boundedContext == "Admin")
-            {
-                loggerConfiguration = new LoggerConfiguration()
-                    .MinimumLevel.Warning()
-                    .Enrich.FromLogContext()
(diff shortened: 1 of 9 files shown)


Source/Infrastructure/AspNet/Initialization.cs
```diff
@@ -30,16 +30,6 @@ public class Initialization
                 .Enrich.FromLogContext()
                 .WriteTo.JsonConsole();
 
-            //TODO: This seemed to fix the problem where Amin was the only BC that would not startup because of an infinite logging loop.
```
karolikl: I assume the infinite logging loop has been fixed by another PR so that we don't end up in a loop again?

woksin: Well, kinda not. I was at Sandefjord trying to figure this out with @smithmx @einari.

I think we landed on that this was just a problem for Visual Studio, we could not reproduce the error by running it in console with Visual Code.

einari: Add to this; it seemed to only happen during debugging as well. The infinite loop is not infinite either - after 4-5 minutes it actually continued.. 

And the cool part; it is totally unrelated to logging. It is during gathering of libraries for discovery mechanisms.

woksin: Also, to add to this. It does not happen to every BC, just Admin, and sometimes User Management

karolikl: @woksin, can you create an issue for the VS debugging issue so that we have it for the future?

PR #730, opened 2018-06-13
Dev fix

woksin:
This should, maybe fix the problems we are seeing in the dev environment where datacollectors and case reports does not show up.

- Puts all extra fields of a ReadModel implementing IHaveExtraElements into ExtraElements
- Creates the interfaces for a quite naive migration system (not in use)


woksin:
Currently this removes the problem where case reports and data collectors are not fetched.

But for case reports we are missing some data from case reports of an older version, I will look into how we can migrate the documents

Source/Infrastructure/Read/MongoDb/IHaveExtraElements.cs
```diff
@@ -0,0 +1,16 @@
...
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization.Attributes;
+
+namespace Infrastructure.Read.MongoDb
+{
+    /// <summary>
+    /// Defines a Read Model class that has extra bson documents.
+    /// ExtraElements will be a dumping place for all extra bson elements when serialized
+    /// </summary>
+    public interface IHaveExtraElements
```
Velcrow81: Do you have examples for when this is bredde? Why not use explicit typed properties?

Source/UserManagement/Read/DataCollectors/Migration/RemoveNationalSociety.cs
```diff
@@ -0,0 +1,19 @@
+using Infrastructure.Read.Migration;
+
+namespace Read.DataCollectors.Migration
+{
```
Velcrow81: Why do we need migration on a read model? Why not just remove property?

Source/Infrastructure/Read/MongoDb/IHaveExtraElements.cs
woksin: ExtraElements is a dump-place for extra elements on the read model. If we have a readmodel with fields that are not present in each of the documents in the mongo collection the application crashes

Source/UserManagement/Read/DataCollectors/Migration/RemoveNationalSociety.cs
woksin: We might have changed the name of field that may or may not be of value to us, we might want to adopt a migration strategy in that case. 

But when we have an event store in place we could just rewind the events and update the read models in that way

Velcrow81: Sounds good. When the event store and replay is available, this scenario will not be necessary longer, although we can still solve it like this when that is easiest/best

Issue #644, opened 2018-04-13
Store 'registered by' when registering a data collector


karolikl:
We need to store who a data collector was registered by.

cathinkaw:
After disccusions with @aninek and @RFMoore we have discovered that they are rather interested in "who is responsible for this volunteer". That is the data verifier, and in most cases the data verifier is the one registering the data collector.

The thougth is therefore that upon registration of a data collector  the data verifier needs to be provided, and if it's possible to get the "person id" from the logged in user, this will be set as the default data verifier.

PR #740, opened 2018-06-20
Issue 644: Add RegisteredBy information to DataCollector classes in backend

cathinkaw:
https://github.com/IFRCGo/cbs/issues/644

@einari could you or anyone else at Dolittle help out with implementing fetching the users guid from the login information? We weren't able to do this as the login is mocked in dev.

In our opinion the fetching of the user information should be done backend.

PS. Please correct our code if we have added RegisteredBy in too many places. We weren't to sure about where it was needed.

### DIFF
--- a/Source/UserManagement/Domain/DataCollector/DataCollector.cs
+++ b/Source/UserManagement/Domain/DataCollector/DataCollector.cs
@@ -3,7 +3,6 @@ using System.Collections.Generic;
 using System.Linq;
 using Concepts;
 using Dolittle.Domain;
-using Domain.DataCollector.Changing;
 using Domain.DataCollector.Registering;
 using Events.DataCollector;
 
@@ -24,7 +23,8 @@ namespace Domain.DataCollector
             string fullName, string displayName,
             int yearOfBirth, Sex sex, Language preferredLanguage,
             Location gpsLocation, IEnumerable<string> phoneNumbers, DateTimeOffset registeredAt,
-            string region, string district
+            string region, string district,
+            Guid registeredBy
             )
         {
             if (_isRegistered)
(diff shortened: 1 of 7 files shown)


einari:
I have a couple of questions:

- Is this for audit trails? The issue seems very technical, what is it going to be used for?
- Is there a relationship between a CBS concept and the user? e.g. DataOwner ? 

The oid that sits on for instance the JWT token coming back is going to be the oid coming from the actual provider. There is no guarantee that this will be a GUID - as different providers have different unique identifiers. In addition to this, in the future the users could be sourced from multiple places. In the federated gateway that Dolittle provides, we will be keeping a GUID to identify them - but I do suspect we probably don't want that coupling.

cathinkaw:
@karolikl or @aninek can you answer Einar's questions?

aninek:
Data collectors should be registered by 'their' data verifiers, so knowing 'registered by' means we know the hierarchy of the CBS design.

einari:
Would it then be fair to say that Data Verifier should be formalized as a concept and with it there is associated an identity and then it becomes the DataVerifier we want to know about. 

From what I can read then is that you're not really that interested in the "RegisteredBy" but more who the data collector belongs to? This is just a convention you'll put in place.  If so, wouldn't be better then to call it "OwnedBy" ?  Or something that refers to the relationship, rather than registered - as that sounds like an auditing concern.

eprom:
@einari I am with you on inconsistent id's coming from  a different identity providers. We do expect a token that the CBS services consume after a user is successfully authenticated. The token should contain a user id and a role that is consistent across the services. Since we are re-signing the token after a user authenticated by IDP, is there any challenge making a consistent user id available to the CBS services? I am aware of we are not there yet on the authorisation :-)

cathinkaw:
It has been decided that what is interesting is the responsible data verifier (https://github.com/IFRCGo/cbs/pull/819). This PR is therefore no longer relevant.

PR #743, opened 2018-06-20
data collecter validation rules

woksin:
Implementet business validation rules for registration and editing of data collectors


Source/UserManagement/Domain/DataCollector/Changing/ChangeBaseInformationBusinessValidator.cs
```diff
@@ -0,0 +1,28 @@
...
+        public ChangeBaseInformationBusinessValidator(IDataCollectorRules dataCollectorRules)
+        {
+            _dataCollectorRules = dataCollectorRules;
+
+            RuleFor(_ => _.DataCollectorId)
+                .Must(BeRegistered).WithMessage(_ => $"Data Collector with id {_.DataCollectorId.Value} is not registered");
+
+            RuleFor(command =>  command) // New display name must not be taken
+                .Must(_dataCollectorRules.DataCollectorCanChangeDisplayName)
+                .WithMessage(_ => $"Datacollector display name {_.DisplayName} is already taken, choose another");
```
Velcrow81: Why can't two data collectors have the same display name? If two people are called John, they should both be allowed to be called that? We have other pices of information that can help differentiate like phone number and location

Source/UserManagement/Domain/DataCollector/Changing/ChangeBaseInformationBusinessValidator.cs
```diff
@@ -0,0 +1,28 @@
+using Concepts.DataCollector;
+using Dolittle.Commands.Validation;
+using FluentValidation;
+
+namespace Domain.DataCollector.Changing
+{
+    public class ChangeBaseInformationBusinessValidator : CommandBusinessValidatorFor<ChangeBaseInformation>
```
Velcrow81: This shounds very generic. Can't the command be ChangeDisplayName?

Source/UserManagement/Domain/DataCollector/Changing/ChangeBaseInformationBusinessValidator.cs
woksin: @aninek should have more input into this. If we allow to have multiple data collectors with the same display name we need to have another way of identyfying data collectors in the case report list (which we can)

PR #746, opened 2018-07-25
Fix error with persisting Concepts 

woksin:


Source/UserManagement/Read/DataCollectors/DataCollectorEventProcessor.cs
```diff
@@ -38,7 +39,7 @@ public void Process(DataCollectorRegistered @event)
 
         public void Process(DataCollectorUserInformationChanged @event)
```
smithmx: Since the cast to the ConceptAs<> type is important and it's repeated over and over, we should probably start a new convention and make the filter part of the mongo query a constant.  Either just a constant lambda that you use or maybe an extension of the mongo collection.

woksin: Yes, we'll have to do that later on

PR #752, opened 2018-08-15
sshpk NPM package updated 

eprom:
The sshpk NPM package is vulnerable to ReDoS when parsing crafted invalid public keys. It is updated from 1.13.1 to 13.2


eprom:
@woksin there are some build fails, do you think this update creating a problem?

PR #772, opened 2018-09-25
Massive cleanup of Admin - feature wise there are a couple of things needing debugging

einari:
We're looking at event processors and why they're not getting called.  Continue this tomorrow morning.


jakhog:
I'm a bit worried that the build fails on Travis and AppVeyor (https://ci.appveyor.com/project/karolikl/cbs-0ntrh/build/1.1.0-63-wggpfssn), but not on Docker Cloud. Are we not running Tests in the Dockerfiles?

Also, just to let you know, the AppVeyor setup is a mess here now, so even though it says OK, the build for Admin failed. It's just showing the result from one of the builds (at random it seems).

woksin:
Admin/Web.Angular fails to build the docker image, @einari

jakhog:
Admin/Web also fails, but not on Docker Cloud...

jakhog:
Could it be as simple as adding:
```
RUN dotnet test
```
In all the Dockerfiles @woksin ?

woksin:
I don't know, I've not played around with it, and neither do I have the time right now

einari:
I'll fix the tests tomorrow. I just had to push it to see if the Angular stuff would work.

PR #786, opened 2018-09-27
Geodesics binding

woksin:


woksin:
I'm doing a local docker build of these projects with no problems

PR #806, opened 2018-09-28
Added paging support in Volunteer Reporting backend

### LINKED ISSUE #792: Implement paging or partial-loading of case reports
The case report list is very slow to load as it retrieves every single case report from the database on every page load. We need to introduce paging or partial-loading to retrieve only a subset of reports at the time.

esp0:
First part of fixing #792


ssokurenko:
Would be good to start the pages count  (PageNumber) from 0.
So the "pageNumber": 1 is the initial page from the beginning of the collection.

woksin:
I'm approving this because this is a temporary hack and is something that we'll implement in the platform later on

PR #819, opened 2018-09-29
Set data verifier for Data Collector upon creation

cathinkaw:
Fixes the backend part of #644.

Because of the possible merge into volunteer reporting/react rewrite/proxy changes we haven't fixed the following:
- Listing all data verfiers in a dropdown list that you can choose from
- Set logged in user as the default data verifier
- Add this in the edit view

### DIFF
--- a/Source/UserManagement/Domain/DataCollector/DataVerifier/ChangeDataVerifierValidator.cs
+++ b/Source/UserManagement/Domain/DataCollector/DataVerifier/ChangeDataVerifierValidator.cs
@@ -1,0 +1,20 @@
+using Concepts.DataCollector;
+using Concepts.DataVerifier;
+using Dolittle.Commands.Validation;
+using FluentValidation;
+
+namespace Domain.DataCollector.DataVerifier
+{
+    public class ChangeDataVerifierValidator : CommandInputValidatorFor<ChangeDataVerifier>
+    {
+        public ChangeDataVerifierValidator()
+        {
+            RuleFor(_ => _.DataCollectorId)
+                .NotEmpty().WithMessage("Data Collector Id must be set")
+                .SetValidator(new DataCollectorIdValidator());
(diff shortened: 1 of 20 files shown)


Source/UserManagement/Domain/DataCollector/DataVerifier/ChangeDataVerifierValidator.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Domain.DataCollector.DataVerifier
+{
+    public class ChangeDataVerifierValidator : CommandInputValidatorFor<ChangeDataVerifier>
+    {
+        public ChangeDataVerifierValidator()
+        {
+            RuleFor(_ => _.DataCollectorId)
+                .NotEmpty().WithMessage("Data Collector Id must be set")
+                .SetValidator(new DataCollectorIdValidator());
+            //RuleFor(_ => _.DataVerifierId)
```
cathinkaw: Outcommented as the serialization of the DataVerifierId (concept as) didn't work. Kept the code here as we should use a DataVerifierId and not a Guid when it works

einari: Could you add an Issue describing the issue [here](https://github.com/dolittle-extensions/ReadModels.MongoDB/issues) with a reference to this file and how we can reproduce it ??

cathinkaw: @woksin can you make the issue? You know more about the details than I.

cathinkaw:
Well spotted @esp0. I have closed it as it is no longer relevant.

einari:
This PR will be part of the #832 PR I have going. We merged it all into that as there was need to clean up user management and make it actually run before we merge it in with volunteer reporting into the new Reporting bounded context

PR #832, opened 2018-09-30
Bringing back User Management to a compilable and runnable state before merging with Volunteer Reporting

einari:
UserManagement builds and runs - but it does not function properly yet. Something that will be fixed very soon - we need to bring this into the new Reporting bounded context and continue working on it from React rather than trying to keep code we're going to drop very soon up to date.


jakhog:
I've changed the Azure DevOps builds to use the same build configuration as the Docker Cloud does, namely building the frontend with the `--prod` flag. I can't figure out how to re-build this PR with the new configuration, but it fails with the same error (local build with `--prod` fails with the same):
```
ERROR in main.e491ebd8e21c1b980c07.bundle.js from UglifyJs
Unexpected token: name (DataCollectorById) [main.e491ebd8e21c1b980c07.bundle.js:62,6]
```

Some digging leads me to believe that this is related to and old version of angular (cli?), which results in an old version of Uglify, that doesn't support the esxxxx syntax used in the proxies. So an update of angular might solve the issue.

Also, there are two copies of the generated proxies `AllDataCollectors.js` and `DataCollectorById.js` which is probably a bad idea.

einari:
You're absolutely right @jakhog - the version of the Angular CLI depends on a version of UglifyJS on UserManagement didn't have support for ESNext syntax - which is what is used in the proxies. Admin was already bumped a few versions, which is why we didn't see this problem there.

einari:
Much appreciated - will you do me the honour of merging it in :)

Velcrow81:
Merged 😊

jakhog:
This PR needs to be celebrated 🍾👏

Issue #843, opened 2018-10-30
HealthRiskById - Object of type 'Infrastructure.Read.MongoDb.ConceptSerializer' cannot be converted to type 'MongoDB.Bson.Serialization.IBsonSerializer`1[Concepts.HealthRisks.HealthRiskId]'."


Nyranith:
When I try to execute the query HealthRiskById I’m currently getting an error with the ConceptSerializer, anyone has any idea why the query fails?

```
{  
   "correlationId":"98d4bfda-b326-a455-bec5-ddf64784af91",
   "nameOfQuery":"HealthRiskById",
   "generatedFrom":"Read.HealthRisks.HealthRiskById",
   "parameters":{  
      "readModel":null,
      "healthRiskId":"25b0a2a5-4a02-46e1-8c0c-885ba40b22fd"
   }
}

```
```

async function getHealthRiskById(healthRiskId) {
    console.warn(healthRiskId);

    var query = new HealthRiskById(); 
    query.healthRiskId = healthRiskId;

    var result = await new QueryCoordinator().execute(query); 

    console.warn(result); 

    if(result.success) {
        return result.items[0]; 
    } else {
        return null; 
    }

}

```

```
{  
   "queryName":"HealthRiskById",
   "totalItems":0,
   "items":null,
   "exception":{  
      "message":"Object of type 'Infrastructure.Read.MongoDb.ConceptSerializer' cannot be converted to type 'MongoDB.Bson.Serialization.IBsonSerializer`1[Concepts.HealthRisks.HealthRiskId]'.",
[...]

PR #844, opened 2018-11-07
Starting on the rewrite on the admin page from angular to react. 

Nyranith:
This is a start on the rewrite to react on the admin page. 

Some notes, the project is created with the command npx create-react-app my-app and then run the eject to get some additional functionality and some more control. 

The pattern used is the _render props pattern_ and after some back and from with a friend we decided to use jsx files endings on all the react components, the non react stuff is having an separate folder called js. 

To run this project first build the cbs.navigation project with 'npm install' and 'npm run build', if we are going to use this kind of project this should probably be moved to a npm repository or added the the react build script. Then go to the Web.React and run the project with as expected with the react commands. 

I also moved all the auto generated functionality from the dolittle web project in a separate folder, since these are removed and auto generated on each build from the web project. The files is not something you should change up in and it can be confusing for new people to use the project. Then you also clearly see what is auto generated and not.  

Currently missing functionality : 
- Post,edit,delete healthcard 
- Not started on the project page yet. 
ETC : Prob Sunday. 

There is also a fix for issue #843 here from @woksin


Nyranith:
@jakhog, could you help me with the docker file?

beatfactor:
@Nyranith I have started already to migrate the Reporting module, you can have a look here: https://github.com/IFRCGo/cbs/tree/master/Source/VolunteerReporting/Web.React. Maybe we could use the same structure for the Admin as well?

karolikl:
Core team has decided not to rewrite Angular frontends that now work. Closing this pull request.

PR #888, opened 2019-01-13
Set correct build mode for Docker Cloud builds that build for dev-cluster

jakhog:
To enable access without login


jakhog:
Pretty confident about this one, so I will merge it without any review.

PR #889, opened 2019-01-13
Added a basic nodejs express server

beatfactor:
Added `Source/Navigation/Web.NodeJS` basic express server to be used by individual React based bounded contexts, primarily for testing and front-end development with a mocked backend.

### DIFF
--- a/Source/Navigation/Web.NodeJS/.eslintignore
+++ b/Source/Navigation/Web.NodeJS/.eslintignore
@@ -1,0 +1,8 @@
+node_modules
+public
+*.html
+*.log
+*.json
+*.md
+*.yaml
+*.ejs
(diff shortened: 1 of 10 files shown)


eprom:
I support to keep it as backup if the backend doesn't work, just in case. We shouldn't hold back the frontend development.

beatfactor:
Mocking front-end code directly is a bit clumsy, in my experience, because you have to change a lot of code usually to do the mocking. It will also be used by the end-to-end tests to fake the backend calls.

Also, an important function can be to perform the authentication and authorization (using the client-credentials flow instead of the implicit flow, which has several shortcomings) and possibly to proxy some of the calls to the backend.

PR #890, opened 2019-01-13
Added React-based components for Reporting bounded context

beatfactor:
Depends on #887 and #889.

This PR adds VolunteersReporting React-based frontend. It is using for now a fake Node.js API server as the backend (which is also used for end-to-end tests).

It has support for writing end-to-end tests, however tests will need to be written.


karolikl:
We'll need to reconfigure the front-end builds in VSTS, the current configuration will keep on failing as it's trying to pull in a bunch of Angular packages. 

@beatfactor, could we look at the build configuration together during the codeathon?

beatfactor:
Yes, I need to get the latest from master in this branch and then I think it should be ok. I'll try and add Dockerfiles but at this point they wont do much besides linting

beatfactor:
@karolikl is the travis build still needed, if we have Azure Pipelines?

PR #892, opened 2019-01-14
Enable multiple tenants for dev setups + Add initial NotificationGateway BC + cross BC event passing

jakhog:
These changes add three other tenants to the configuration for Docker Compose and Kubernetes (running in the dev-cluster). The tenant is set by the sub-domain of the request. So navigating to either URLs like:
- admin.*
- country_one.*
- country_two.*
Will get you into three different tenants, simulating multiple deployments.


jakhog:
Ooops 🙈 Pushed some other changes to my master, and forgot about this PR. It now contains a lot more than originally planned.

jakhog:
So, in addition to the original PR contents, these changes now also contain an initial version of the NotificationGateway bounded context.

This BC contains only a single command to simulate the reception of an SMS, which triggers the `TextMessageReceived` event. It has a very simple GUI setup to "receive" and SMS, and show all previous SMSes.

Also, the VolunterReporting BC have been set up to subscribe to this event, and has gotten an EventProcessor that simply writes warnings to the log that this event should be processed.

jakhog:
The purpose of the NotificationGateway code as it stands now, was to showcase how inter-BC events work (and that they work), provide a framework for the work on the Gateway at the upcoming Codeathon, and lastly to provide a simple method to do end-to-end testing of the system in the dev cluster without setting it up with the SMSeagles and using actual phones.

Automated Docker builds will be set up once the PR is merged, and then it will also be deployed to the dev-cluster.

PR #899, opened 2019-01-16
Merge 'VolunteerReporting' and 'UserManagement' backends into 'Reporting'

### LINKED ISSUE #896: Merge Volunteer Reporting and User Management backends
They will form the new Bounded Context 'Reporting'

woksin:
The main work of migrating the VolunteerReporting and UserManagement backends over to the new Reporting bounded context should be done with this PR. This is basically just a copy paste in addition to changing some names and making it fit a 'Module' based feature system.

Though the hard work, in my opinion, still stands - moving the frontends over...

I think that it's important that we manage to get this in by the Codeathon because the work that we're going to do on 'Reporting' now should be done in this new bounded context so that we don't need to move parts over again.

Since this is basically is a copy/paste we might stumble over things to improve, most likely. There is, for example, a lot of improvements we can do on the read models in terms of simplifications and performance improvements.

@jakhog After we've pulled this in we need to set up the deployment steps for this BC as well as including it in our CI

fixes #896


beatfactor:
would it be possible to squash the commits?

woksin:
Waiting for @jakhog to finish this up

Issue #834, opened 2018-10-01
Split incoming case reports into cases


Velcrow81:
Add policy that listens for Case report recieved
Create aggregate root called Case
Emit event Case registered
Create read model for cases and add case to this

Ask Velcrow81 or Aninek when you have questions

Velcrow81:
Overview of alert events
https://realtimeboard.com/welcomeonboard/PnBU5ceLe67kUqNSNJi2nxwO3U70g6vtbMCfSyNLcnaMyeoU8WBABDxigjZFgIWv

PR #914, opened 2019-01-18
Handle report received

bullmastiffo:
Implements splitting CaseReport into separate Cases #834


bullmastiffo:
> The name of the projects in the bounded context solution should not be prefixed by the BC name. So `Alerts.Concepts` -> `Concepts`, and `Alerts.Events` -> `Events`.
> 
> Also, the guid of the `Events.CaseReports.CaseReportReceived` was changed. Which means that it is a whole new event, so we can't do that.

@jakhog Thanks, I've addressed the issues, please have a look

Issue #926, opened 2019-01-19
MVP Backend: Send notification to data owner when alert is triggered


mfredbo:
Once an alarm is triggered, send a notification to the Data owner on e-mail.

For now, the content in the email should be hard coded. The intention is that at least the data owner will get a notification in their mailbox.

Issue #923, opened 2019-01-19
MVP Frontend + UI: Make a page where all the alert rules are listed


mfredbo:
**Description**
This page should include a list off all registered alert rules. It should also be possible to open an alert rule and look at it.

Editing and deleting alert rules will be a feature implemented later on.

gryha:
Here's the link to user interface in UXpin: https://preview.uxpin.com/cdf72b60c80d94cf32507cadfaa0b1888f3d66f2#/pages/101874840

Issue #920, opened 2019-01-19
MVP Backend - An alert rule is triggered based on one case registered


mfredbo:
**MVP - Alert rule is triggered**
Once one _case is registered_ (on one health risk), I want an alarm to be triggered. This is backend.

This is the MVP. 

I.e., later on, the alarm will later on be triggered based on also threshold (such as number of cases, number of hours between the cases registered, and distance between cases registered).

Please define this task more detailed if needed, or ask me if you have any questions.

Velcrow81:
We improved it to be threshold and time dependant. Distance is ignored for now

Issue #912, opened 2019-01-18
As a data owner, I want to be able to register an alert rule - Backend


mfredbo:
**This is linked to issue number 906**
This issue is the backend part of the issue here: https://github.com/IFRCGo/cbs/issues/906
This is the issue that  includes the tasks needed to make the "alert rule" function. This issue might need to be detailed into several smaller tasks.

**I want the following to be implemented**:
- [ ]  Health risk name : There should be a look up table with names of health risks, so that the user can choose the health risks from a list. This list might be an input we will get from the Admin team later on, but for now, we will implement this ourself.
- [ ] Cases: This field represents the number of cases which should trigger an event. The number that is submitted here should be registered/stored as the threshold for when a new alert should be triggered.
- [ ] I want to count cases...: The input here should be stored as the timeframe for when the threshold is valid in between. 
- [ ] Distance: This is the distance defined where similar cases within one health risk should be aggregated. When the cases within this distance is aggregated to the case threshold, and within the timeframe given, then an alert should be triggered. Check with @raubreywhite how the distance should be  interpreted.

mfredbo:
<img width="638" alt="alert rule updated dec 18 2019" src="https://user-images.githubusercontent.com/20842471/51395134-6780e900-1b3c-11e9-88d3-11f0e3ba8f2f.png">

PR #948, opened 2019-01-19
Alert rule triggering SMTP send

bullmastiffo:
Tasks implemented  #906 #920, #923, #912 #926


Velcrow81:
@jakhog : Do you understand why docker fails for NotificationGateway? No changes should affect it in this bounded context

jakhog:
@Velcrow81 the Docker builds are not behaving nicely all the time, it just failed for no reason at all :/ All green now

PR #954, opened 2019-01-21
Reporting - merge in Specs and validation

woksin:
Most of this stuff is made by @kaithelin, I just had to copy it over manually because of some horrible git merging problems...

Implements lots of validation on commands and specs. 

I would want to test this together with the old UserManagement front end to see if the validation works correctly, but I cannot seem to get the old angular frontend working with this backend due to apparently CORS problems? cc @jakhog


woksin:
It does not look like CI for the Reporting project is included here? @jakhog

woksin:
I'll sneak in some more stuff here that I saw needed a fix...

jakhog:
The Docker builds were failing because I made a mistake while deploying the AspNetCore Dolittle packages, so they should pass now. I just restarted them so we will se. The Azure CI builds also decided to start when I went to look at them, I have no idea why they didn't start before 🤷‍♂️ It's like Schrödingers cat...

PR #978, opened 2019-02-11
Refactored from Case to Report in Alert. Also turned on proxy generation again

Velcrow81:


Velcrow81:
1. The renaming is actually from case to report. This is only related to the Alert bounded context. In Reporting, it is still called CaseReport, but that should also be changed. It might be an issue with both beeing called Report, but thats the beauty of bounded contexts. Something can be called the same but be different things in different bounded contexts. 

2. Reports have always been related to human and non-human health risks. Flood, fire and unusual events are examples of non-human health risks, but they are still health risks since people can die from it. There are special health risk id numbers for such helath risks, but they are reported with only a health risk id or with 0 for all values related to sex and age

3. CaseReportRecieved is an event from Reporting and since that has not been renamed yet, it is not changed here. The reason we split it into many Reports is because one CaseReport can contain multiple events (aggragates). I.e. One CaseReport can contain 10 girls under the age of 5. They are split into individual reports since they will be followed up individually by a supervisor, and dismissed or keept based on a phone call on symptoms. Then each alert report is counted to see if any thresholds are crossed. Make sense?

PR #985, opened 2019-03-06
Add backend-endpoint for excel export

### LINKED ISSUE #971: Create backend endpoint for excel-export
We need an endpoint in the backend to have something to call when the Export to Excel button is clicked. This is a requirement for #967

esp0:
Fixes #971


jakhog:
I have copied the work here into #1000, so no need for this one. Still good work though @esp0 😄

PR #1053, opened 2019-04-13
Dakar alert codathon result

Velcrow81:
- Done work to make it easier to work on frontend. Added mocks that will isolate frontend from backend so .net and docker is not needed. Can also use start script that use local api or demo api

- Added new read model for alert overview and connected this to alert opened event. Also created mock test data for all existing read models

- Added static UI for alert overview, alert rule overview, and data owner with navigation. Mock data/ real data not used yet

- Missing top navigation after converting to Notification gateway React template


Velcrow81:
Waiting for one last commit, so no merging yet :)

PR #1067, opened 2019-04-18
 Add multiselect filter for location in case report list - attempt 2

woksin:
Attempt number 2 of this PR :) 

I reverted the changes made when this Pull Request was pulled in before, it deleted changes made for globalization and other stuff that we want.

I made the frontend build by switching out the mulit-select component with "angular2-multiselect-dropdown", so use that instead of the one used before.

I'm sorry for the trouble, but it would be really nice to have this in feature in the app :)


woksin:
Woops, wrong repo to make PR to, my bad

Issue #1070, opened 2019-04-22
Analytics does not run locally after adding application insights logging


karolikl:
After PR #1064, the Analytics bounded context does not run locally. The error I get is this: 
![image](https://user-images.githubusercontent.com/5235776/56488339-674b0d80-64de-11e9-884d-32f1e9a7ef38.png)

If i comment out the following (in Program.cs) it works: 
`     //.ConfigureLogging(builder => {
                //    builder.AddApplicationInsights(Environment.GetEnvironmentVariable("ASPNETCORE_APPINSIGHTS_INSTRUMENTATIONKEY"));
                //    builder.AddFilter<Microsoft.Extensions.Logging.ApplicationInsights.ApplicationInsightsLoggerProvider>
                //                    ("Analytics.Program", LogLevel.Trace);
                //    builder.AddFilter<Microsoft.Extensions.Logging.ApplicationInsights.ApplicationInsightsLoggerProvider>
                //                    ("Analytics.Startup", LogLevel.Trace);
                //})`

Are we missing ASPNETCORE_APPINSIGHTS_INSTRUMENTATIONKEY?

gardnk:
The ASPNETCORE_APPINSIGHTS_INSTRUMENTATIONKEY is an environment variable so it needs to be available locally. I'm running from Visual Studio Code and have it configured in the launch.json like this

```
        {
            "name": ".NET Core Launch (web)",
            "type": "coreclr",
            "request": "launch",
            "preLaunchTask": "build",
            "program": "${workspaceFolder}/Source/NotificationGateway/Core/bin/Debug/netcoreapp2.1/Core.dll",
            "args": [],
            "cwd": "${workspaceFolder}/Source/NotificationGateway/Core",
            "stopAtEntry": false,
            "launchBrowser": {
                "enabled": true
            },
            "env": {
                "ASPNETCORE_ENVIRONMENT": "Development",
                "ASPNETCORE_APPINSIGHTS_INSTRUMENTATIONKEY": "36bdf7e0-884f-4391-8f08-11ebd48b9023"
            },
            "sourceFileMap": {
                "/Views": "${workspaceFolder}/Views"
            }
        }
```

gardnk:
Now it works when running locally using docker-compose also. All that is needed is having ASPNETCORE_APPINSIGHTS_INSTRUMENTATIONKEY configured on your host machine.

gardnk:
[...]

PR #1074, opened 2019-04-25
the projects don't crash when app insights instrumentationkey not pre…

gardnk:
Now checking if the environment variable is present before initializing application insights to prevent crashing.


esp0:
This closes #1070 i think?

Issue #1043, opened 2019-04-13
Rename files that are too long


cathinkaw:
There are some files in the solution that gives people a "file name too long" error on Windows. This can be solved with config, but I think it is not necessary to have files with so long names either.

cathinkaw:
'Source/Reporting/Policies.Specs/Reporting/for_notifications/when_parsing_a_text_message_with_a_correctly_formated_aggregated_case_report/separated_by_hash/given/text_message_received_events_containing_aggregated_case_report_separated_by_hash.cs': Filename too long
error: cannot stat 'Source/Reporting/Policies.Specs/Reporting/for_notifications/when_parsing_a_text_message_with_a_correctly_formated_aggregated_case_report/separated_by_star/given/text_message_received_events_containing_aggregated_case_report_separated_by_star.cs': Filename too long
error: cannot stat 'Source/Reporting/Policies.Specs/Reporting/for_notifications/when_parsing_a_text_message_with_a_correctly_formated_aggregated_case_report/separated_by_star_and_hash/given/text_message_received_events_containing_aggregated_case_report_separated_by_star_and_hash.cs': Filename too long
error: cannot stat 'Source/Reporting/Policies.Specs/Reporting/for_notifications/when_parsing_a_text_message_with_a_correctly_formated_aggregated_case_report/separated_by_star_and_hash/when_report_has_a_negative_number_of_cases_of_females_over_5.cs': Filename too long
error: cannot stat 'Source/Reporting/Policies.Specs/Reporting/for_notifications/when_parsing_a_text_message_with_a_correctly_formated_aggregated_case_report/separated_by_star_and_hash/when_report_has_a_negative_number_of_cases_of_females_under_5.cs': Filename too long
error: cannot stat 'Source/Reporting/Policies.Specs/Reporting/for_notifications/when_parsing_a_text_message_with_a_correctly_formated_aggregated_case_report/separated_by_star_and_hash/when_report_has_a_negative_number_of_cases_of_males_over_5.cs': Filename too long
error: cannot stat 'Source/Reporting/Policies.Specs/Reporting/for_notifications/when_parsing_a_text_message_with_a_correctly_formated_aggregated_case_report/separated_by_star_and_hash/when_report_has_a_negative_number_of_cases_of_males_under_5.cs': Filename too long
[...]

PR #1087, opened 2019-05-14
Solve problem with too long file names on Windows

cathinkaw:
Files arent possible to open in VS on Windows and you have to change git config to download the repo. 

Solves this: https://github.com/IFRCGo/cbs/issues/1043


cathinkaw:
@gardnk can you look at this instead maybe?

gardnk:
@karolikl or @Velcrow81 can you test this? I have a mac so that would defeat the purpose 😅

PR #1088, opened 2019-05-14
Add mocking for Reporting

cathinkaw:
Setup to be able to work independantly on frontend in Senegal.

Can this be merged as is @woksin to get a smaller diff on the rest of the code?


cathinkaw:
Rather updating with my own branch based on Sindres repo, than his.

Issue #1054, opened 2019-04-13
Review code for alert rule and alert menu and decide how to merge it


Velcrow81:
Code added in Dakar was done in parallell to some existing code. Pick the best from both and merge.

PR #1093, opened 2019-05-22
Merging work on alerts from Dakar

### LINKED ISSUE #1051: Connect alert overview, alert rule and data owner with coordinators and remove static html test data
Then we can use mock data for local UI development or get data from demo environment

### LINKED ISSUE #1055: Alert menu should be moved from alert rule folder and become a sepperate component

### LINKED ISSUE #1066: Remove all inline styling from Alerts frontend
We don't want inline styling in HTML on our components, styles should be stored in seperate .scss-files or another equivalent way of keeping it in one place.

gardnk:
* Components made in Dakar now uses coordinators fixes #1051 
* Static HTML data is removed
* Inline-styling removed fixes #1066 
* Alert menu component renamed and moved out of the folder fixes #1055 

This is in progress as there's still some work to be done in merging the components that are created in parallell to existing code.


gardnk:
In my mind the only difference between what was already done and what was done in Dakar is the use of Evergreen vs Material-UI.

I would personally go with Evergreen as its default styles are what the design is made for. To me this looks like less trouble implementing.

@Velcrow81 any thoughts?

Velcrow81:
Sounds reasonable to me :)

gardnk:
Last commit does the final merging - fixes #1054

PR #1095, opened 2019-05-23
Proxy fixes

gardnk:
I updated two javascript commands to use ```number``` and not ```{}```

Also changed enum values from string to number so that it builds, but I'm not sure what numbers to use so need some help there @woksin please and thank you


Source/Reporting/Web/src/app/Reporting/CaseReportsForListing/filtering/filter.pipe.ts
```diff
@@ -2,10 +2,10 @@ import {Pipe, PipeTransform} from '@angular/core';
 import {CaseReportForListing} from '../CaseReportForListing';
 
 enum CaseReportStatus {
-    Success = "Success",
-    TextMessageParsingError = "TextMessageParsingError",
-    UnknownDataCollector = "UnknownDataCollector",
-    TextMessageParsingErrorAndUnknownDataCollector = "TextMessageParsingErrorAndUnknownDataCollector"
+    Success = 0,
```
gardnk: I don't know what numbers to use here 🤷‍♂

woksin: The enum value definitions can be found in the backend c# code with the same names

Look in the Concepts project and the Readmodels project for enum definitions

gardnk: Ok thanks, looks like it was correct

gardnk:
This runs like a charm on my end, can you review @woksin ?

Issue #1114, opened 2019-05-31
[Country overview] Show active and inactive data colletors in map


cathinkaw:
Show active and inactive data collectors for a project on a map. Linked to https://github.com/IFRCGo/cbs/issues/1113 as it shows the same data.

![image](https://user-images.githubusercontent.com/1155078/58712713-6e104e80-83c1-11e9-9e43-c60afb46b5b9.png)

https://github.com/IFRCGo/cbs/wiki/Analytics

agnetedjupvik:
Map coordinates for data collectors seem unavailable for now – I instead mapped the reports. This is lacking because a data collector hasn't necessarily reported anything and will therefore not be seen on the map as per now.

tineml:
do the data collectors need to send one initial report to be able to be seen later in the map?
If possible, it would be good to map all registered data collectors in their village and then mark if they have sent a report. Good if there can be an option to see last 7 days and last 30 days

agnetedjupvik:
At the moment, yes, but I think it's more than possible to include this as information connected to the data collector personally. I made an issue for it #1118, it will probably be addressed this summer:)

Issue #1113, opened 2019-05-31
[Country overview] Show text overview of data collectors


cathinkaw:
Show active data collectors and inactive data collectors for the country as text. 

Some sort of icon may be used to make it easy to recognize (NB. Remember that we should have as little data to download as possible so use ie svg).

![image](https://user-images.githubusercontent.com/1155078/58712504-f7735100-83c0-11e9-8f62-5cf81eda6b4a.png)

https://github.com/IFRCGo/cbs/wiki/Analytics

Issue #1112, opened 2019-05-31
[Country overview] Show total number of case reports


cathinkaw:
The total amount of case reports for the Country should be shown, listing case reports per healt risk underneath. Some sort of icon may be used to make it easy to recognize (NB. Remember that we should have as little data to download as possible so use ie svg).

![image](https://user-images.githubusercontent.com/1155078/58712103-36ed6d80-83c0-11e9-8a32-6c8fd840123d.png)

https://github.com/IFRCGo/cbs/wiki/Analytics

agnetedjupvik:
WIP Branch: [Here](https://github.com/IFRCGo/cbs/compare/master...agnetedjupvik:header-panel?expand=1)
![Screen Shot 2019-06-03 at 4 52 09 PM](https://user-images.githubusercontent.com/10497878/58811468-29362300-8620-11e9-9354-95063230f420.png)

PR #1117, opened 2019-06-04
Project Presence panel

agnetedjupvik:
**Features**
- Displays total number of case reports (solving #1112)
- Displays text overview of active/inactive data collectors (solving #1113)
- Displays map of case reports (partially solving #1114)

**Other changes**
- Renamed AnalyticsBanner to ProjectPresence, as it displays the overall project presence parts of the analytics data
- Introduced HeaderPanel component consisting of an icon, headline and list of relevant information
- Introduced some inline styling. This isn't used anywhere else as of now (there isn't much custom styling at all), but may be a good idea as it loads only the relevant CSS of the components displayed rather than all CSS at once

### DIFF
--- a/Source/Analytics/Web.Frontend/src/components/HeaderPanel.js
+++ b/Source/Analytics/Web.Frontend/src/components/HeaderPanel.js
@@ -4,8 +4,12 @@ import { Heading, UnorderedList, ListItem } from "evergreen-ui";
 export class HeaderPanel extends React.Component {
 
     render() {
+        let panelStyle = {
+            marginRight: 20
+        };
+
         return (
-            <div className="analytics--headerPanel">
+            <div className="analytics--headerPanel" style={panelStyle}>
                 <i className={`${this.props.icon} fa fa-5x analytics--headerPanelIcon`} />
 
                 <Heading color={this.props.color} size={600} paddingTop={"20px"}>
--- a/Source/Analytics/Web.Frontend/src/components/HeaderPanel.js
+++ b/Source/Analytics/Web.Frontend/src/components/HeaderPanel.js
@@ -5,7 +5,7 @@ export class HeaderPanel extends React.Component {
 
     render() {
         let panelStyle = {
-            marginRight: 20
+            marginRight: 30
         };
 
         return (
--- a/Source/Analytics/Web.Frontend/src/components/HeaderPanel.js
+++ b/Source/Analytics/Web.Frontend/src/components/HeaderPanel.js
@@ -1,0 +1,29 @@
+import React from "react";
+import { Heading, UnorderedList, ListItem } from "evergreen-ui";
+
+export class HeaderPanel extends React.Component {
+
+    render() {
+        return (
+            <div className="analytics--headerPanel">
+                <i className={`${this.props.icon} fa fa-5x analytics--headerPanelIcon`} />
+
+                <Heading color={this.props.color} size={600} paddingTop={"20px"}>
+                    {this.props.headline}
+                </Heading>
+
+                <div className="analytics--listContainer">
+                    <UnorderedList size={500}>
+                        {this.props.list.map(
+                            (data, index) => (
+                                <ListItem key={index}>
+                                    {data.numberOfReports || data.inactiveDataCollectors} {data.name || data.healthRisk}
+                                </ListItem>
+                            )
+                        )}
+                    </UnorderedList>
+                </div>
+            </div>
+        );
+    }
+}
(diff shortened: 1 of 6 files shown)


Source/Analytics/Web.Frontend/src/components/HeaderPanel.js
```diff
@@ -0,0 +1,33 @@
...
+                <Heading color={this.props.color} size={600} paddingTop={"20px"}>
+                    {this.props.headline}
+                </Heading>
+
+                <div className="analytics--listContainer">
+                    <UnorderedList size={500}>
+                        {this.props.list.map(
+                            (data, index) => (
+                                <ListItem key={index}>
+                                    {data.numberOfReports || data.inactiveDataCollectors} {data.name || data.healthRisk}
```
cathinkaw: Shouldn't rather the data that we send in be shaped so we have the correct format of it, instead of "forcing" this component to be linked to the datastructure (and name of properties) of the input data?

Source/Analytics/Web.Frontend/src/components/Map.js
```diff
@@ -90,13 +90,15 @@ class MapWidget extends Component {
             );
         }
         return (
-            <Map center={position} zoom={6}>
+            <Map center={position} zoom={5} style={{height: 210, width: 210}}>
+                {/* TODO: Auto zoom and center based on set of markers */}
```
cathinkaw: Can this be moved to an issue instead of a comment?

Source/Analytics/Web.Frontend/src/components/ProjectPresence.js
```diff
@@ -154,65 +154,47 @@ class AnalyticsBanner extends Component {
...
+                        icon="fa-user"
+                    />
+                    
+                    <div>
+                        {/*<Heading size={600} color={"#9f0000"}>Reports</Heading>*/}
+                        <Map />
                     </div>
+                    
+                    
+                    {/*
```
cathinkaw: As this is not being used I think it can rather be removed and readded later if needed.

Source/Analytics/Web.Frontend/src/components/ProjectPresence.js
```diff
@@ -154,65 +154,47 @@ class AnalyticsBanner extends Component {
...
+                    
+                    <HeaderPanel 
+                        headline={`${this.state.dataCollectors.activeDataCollectors} Active Data Collectors`}
+                        list={[{inactiveDataCollectors: this.state.dataCollectors.inactiveDataCollectors, name:"Inactive Data Collectors"}]}
+                        color="#009f00"
+                        icon="fa-user"
+                    />
+                    
+                    <div>
+                        {/*<Heading size={600} color={"#9f0000"}>Reports</Heading>*/}
```
cathinkaw: should rather be removed

cathinkaw:
I also have a question. Back in the days when I last used react we defined the proptypes. Is this something that isn't done anymore? And could maybe my need to know what a component expects be solvec by adding typescript? (Ref: https://github.com/IFRCGo/cbs/issues/1126)

Source/Analytics/Web.Frontend/src/components/HeaderPanel.js
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import { Heading, UnorderedList, ListItem } from "evergreen-ui";
```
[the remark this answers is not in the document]
agnetedjupvik: Yes, planning on taking Evergreen out soon, but I left it out of this PR

Source/Analytics/Web.Frontend/src/components/HeaderPanel.js
```diff
@@ -0,0 +1,33 @@
...
+
+export class HeaderPanel extends React.Component {
```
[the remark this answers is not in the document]
agnetedjupvik: I agree! I renamed it to PresenceIndicator, that fits with its current use, but it can be used for many other things, so I'm open to name suggestions:)

Source/Analytics/Web.Frontend/src/components/HeaderPanel.js
agnetedjupvik: Agreed. I rewrote it, now it takes any JSON Object and lists them "value-first" no matter the keys:

`
{name: "Cholera", numberOfReports: "20"} --> "20 Cholera"
`

Not very scalable, but compatible with the current API endpoints. Could be a good idea to include this as a test in a frontend testing framework in the future.

Source/Analytics/Web.Frontend/src/components/Map.js
agnetedjupvik: Yep!

Source/Analytics/Web.Frontend/src/components/ProjectPresence.js
agnetedjupvik: Agreed

agnetedjupvik:
Yes, I think TypeScript is a great idea. Adding proptypes also work, but TypeScript is stricter, which I think will help streamline different contributions.

PR #1132, opened 2019-06-24
Connect Analytics to Admin events

agnetedjupvik:
Analytics now can listen to Admin events (and later on, events from other bounded contexts). Health risks, coming from Admin, are now modeled as Dolittle artifacts and generate proxies using Dolittle workflow.


gardnk:
Also included in this PR:
* Analytics restructured using Dolittle all the way
* Analytics listening to CaseReportReceived event from Reporting.

PR #1148, opened 2019-07-12
Analytics using Dolittle for all fetching + some spring cleaning

agnetedjupvik:
__New features__
- All data is now fetched into Analytics using Dolittle Queries
- Navigation header has an updated design, now looking like the navigation on the rest of the site (which uses Angular) and looking much better on mobile/small screens 
- Navigation menu now displays Reports in stead of Reporting, Data Collectors instead of User Management
- Alerts are removed from the navigation bar (but still available under /alerts so that things can be tested there – but nothing is up and running now, so it is removed)

__Testing feature__
- For testing, now `Day.Today` gives a day which we knows has data in the database (id 17925). This can easily be reverted, but it's convenient for testing purposes as the database will be reliably populated.

Screen Shot 2019-07-12 at 1 01 18 PM

__How to test__
Run Reporting backend, and in Swagger at localhost:5003/swagger, use the following command:
`
{
  "correlationId": "510ef709-01ce-4255-9459-a70a274bcbe3", 
   "type": "79c674e6-6f04-476a-8608-1ddfcde8f702",
   "content": {}
}
`
Analytics will listen to events and populate reports, data collectors and health risks.


Source/Analytics/Core/Core.csproj
```diff
@@ -43,11 +43,10 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Concepts\Concepts.csproj" />
-    <ProjectReference Include="..\Domain\Domain.csproj" />
     <ProjectReference Include="..\Events.Reporting\Events.Reporting.csproj" />
     <ProjectReference Include="..\Events.Admin\Events.Admin.csproj" />
     <ProjectReference Include="..\Read\Read.csproj" />
-    <ProjectReference Include="..\Policies\Policies.csproj" />
+    <ProjectReference Include="..\Read.Specs\Read.Specs.csproj" />
```
gardnk: I suggest removing this project as it is empty

PR #1150, opened 2019-07-15
Alter test data dates to be pinned to today

cathinkaw:
Instead of some magic to create test data dates, I just reset the "zero" of the dates to today.


cathinkaw:
Can you have another look @gardnk ?

PR #1155, opened 2019-07-19
Casereport map

ojoha:
Clustering map with clustering of health-risks.
If it is multiple health-risks in a cluster, the cluster will be brown.   
Also changed the testdata in reporting for case-reports, so multiple health-risks can appear on the map.


Source/Analytics/Web/src/components/Map.js
```diff
@@ -3,43 +3,113 @@ import { connect } from "react-redux";
...
+var iconsClasses = []
+
+function MarkerPopupContent(props){
+    return <div>{props.healthRisk}</div>
+}
+
+function CaseMarkers({caseReportsLastWeek}){
+    // Returns one Marker for each case in a case-report and sets an unique color for each specific health risk
+    return  Object.keys(caseReportsLastWeek.caseReportsPerHealthRisk).map(function(key) {
+        var healthRiskName   = caseReportsLastWeek.caseReportsPerHealthRisk[key][0].healthRiskName
```
cathinkaw: number of spaces before =

Source/Analytics/Web/src/components/MapOverview.js
```diff
@@ -0,0 +1,35 @@
+import React, { Component } from "react";
+
+function ListOfHealthRisks({clusters,healthRisks, icons}){
+    
+    var listItem = healthRisks.map(function(healthRisk, i){
+        var cl = icons[i];
+        var clo = clusters[i];
+        return <li className="map-overview-list-item"><div className={clo}></div> <i class={cl} ></i> : {healthRisk}</li>
```
gardnk: class in ``` <i class={cl} ></i>``` should be className

Source/Analytics/Web/src/components/Map.js
```diff
@@ -3,43 +3,113 @@ import { connect } from "react-redux";
...
+        healthRisks.push(healthRiskName)
+        iconsClasses.push(markerClass)
+        clusters.push(clusterClass)
+
+        var cs = caseReportsLastWeek.caseReportsPerHealthRisk[key].map(cases => {
+            cases.vals = new Array(cases.numberOfPeople);
+            cases.vals.fill(1);
+            return cases.vals.map(function(val){
+                var hl = healthRisks[healthRisks.length - 1]
+                return <Marker
```
gardnk: Add key to ```<Marker>``` to satisfy React.
I suggest using index in the list, you can get index like this:
```cases.vals.map((val, index) => )```

Source/Analytics/Web/src/components/MapOverview.js
gardnk: Here also use key in the list for React not to complain. You can use the variable i in the map function

Source/Analytics/Web/src/components/Map.js
gardnk: This should also handle the case of no data (caseReportsLastWeek = undefined) so that it doesn't throw an error.

PR #1157, opened 2019-07-22
Redesign header

agnetedjupvik:
The menu is redesigned according to new UX sketches.

The Country Overview is taken out as a separate component. Different pages in Analytics have a submenu for each. Currently only pointing to /analytics – the point here is a dropdown functionality.

On desktop:
Screen Shot 2019-07-22 at 2 58 43 PM

On mobile: 
Screen Shot 2019-07-22 at 2 58 25 PM

On mobile, the hover functionality may give issues with opening the submenu. If anyone has any clever ideas on how we can hover on big screens and click on small screens, that would be appreciated 🎈

Note that this header change is in Analytics (React) only. We should either do a similar rewrite in the other bounded contexts using Angular, or look into using the same component for the header sitewide.


agnetedjupvik:
Fixed your comments now, @cathinkaw. Awaiting reply from Hannah on the "Logged in" info, as it was not included in the sketches (but probably should be).

agnetedjupvik:
Login status is now updated.
Screen Shot 2019-07-25 at 1 21 50 PM
Screen Shot 2019-07-25 at 1 22 06 PM

PR #1158, opened 2019-07-23
Display reports per sex and age in a table

agnetedjupvik:
Now displays reports per sex and age according to new UX sketches. 

The query for this table now returns the report data from the last 7 days for: 
- Females under 5
- Males under 5
- Females over 5
- Males over 5

The totals are calculated using these numbers in the frontend.

Screen Shot 2019-07-23 at 11 00 26 AM


Source/Analytics/Web/src/components/lastWeekTotals/LastWeekTotals.js
```diff
@@ -0,0 +1,101 @@
...
+                    isError: false
+                });
+            }
+        }).catch(_ => {
+            this.setState({
+                isLoading: false,
+                isError: true
+            })
+        }
+        )
```
cathinkaw: Pirk: on same line as } ?

Source/Analytics/Web/src/components/lastWeekTotals/LastWeekTotals.js
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    componentDidMount() {
+        this.fetchLastWeekTotals();
+    }
+
+    render () {
+        var totalUnder5 = parseInt(this.state.femalesUnder5) + parseInt(this.state.malesUnder5);
```
cathinkaw: Is it better to parse the numbers in fetchLastWeekTotals?

agnetedjupvik: Yes! I was worried this would cause a problem as the state starts out as a string `"-"`, but then I remembered that this is JS so it's okay 🐒

agnetedjupvik:
I centered the numbers and gender symbols. Hannah and I agreed that all lines should be of equal width, but the sketches aren't updated yet.

PR #1161, opened 2019-07-26
Reports per health risk per day

gardnk:
Added component and read models to show number of reports per region the last week, filterable on health risk.

According to sketch: https://preview.uxpin.com/6f7c2440d8ba5f7888d63932bbc82c4138712847#/pages/114863485/simulate/sitemap?mode=i

The most recent day is to the right in the table, opposite of the sketch. I got exceptions when reversing the list on the backend so I skipped it for now.


Source/Analytics/Web/src/Features/DataCollectors/DataCollector.js
```diff
@@ -16,5 +16,7 @@ export class DataCollector extends ReadModel
         this.lastActive = new Date();
         this.region = '';
         this.district = '';
+        this.location = {};
+        this.sex = 0;
```
agnetedjupvik: Is it necessary to know the sex of a DataCollector in Analytics?

Source/Analytics/Web/src/components/NationalSocietyOverview.js
```diff
@@ -41,7 +42,8 @@ class NationalSocietyOverview extends Component {
                     </Grid>
 
                     <HealthRiskPerDistrictTable />
-                    <Map />
+                    {/* <Map /> */}
+                    <ReportsPerHealthRiskPerDay></ReportsPerHealthRiskPerDay>
```
agnetedjupvik: As this component has no children, it should be self-closing: `<ReportsPerHealthRiskPerDay />`

Source/Analytics/Web/src/components/Reports/ReportsPerHealthRiskPerDay.js
```diff
@@ -0,0 +1,154 @@
...
+
+
+export default class ReportsPerHealthRiskPerDay extends Component {
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            reportsPerHealthRisk: [],
+            regions: [],
+            healthRisks: [''],
```
agnetedjupvik: Does this need to be initialized with an empty string inside?

Source/Analytics/Web/src/components/Reports/ReportsPerHealthRiskPerDay.js
```diff
@@ -0,0 +1,154 @@
...
+                    return <TableCell className="cell" key={i}>{reportsInRegion}</TableCell>;
+                else
+                    return <TableCell className="cell" key={i}>-</TableCell>;
+            });
+        }
+    }
+
+    renderRegions() {
+        return this.state.regions.map(region => {
+            return (
```
agnetedjupvik: I think you can skip the second `return` statement, by mapping like this: `...region.map(region => (<TableRow>...</TableRow>)`

PR #1167, opened 2019-07-30
Reports per health risk and region last4weeks

gardnk:
* Finished and restructured the component @agnetedjupvik was working on.
* Added to country overview and lightweight overview


Source/Analytics/Read/CaseReports/CaseReportsPerRegionLast4WeeksQuery.cs
```diff
@@ -5,16 +5,16 @@
 
 namespace Read.CaseReports
 {
-    public class CaseReportsPerRegionLast7DaysQuery : IQueryFor<CaseReportsPerRegionLast7Days>
+    public class CaseReportsPerRegionLast7DaysQuery : IQueryFor<CaseReportsPerRegionLast4Weeks>
```
cathinkaw: The name of the query needs to be renamed as well (i tried but couldn't find the documentation for what the dolittle command is aagain..).

Source/Analytics/Web/src/components/Reports/ReportsPerHealthRiskPerRegionLast4Weeks.js
```diff
@@ -0,0 +1,109 @@
+import React, { Component } from "react";
+import Table from '@material-ui/core/Table';
+import TableBody from '@material-ui/core/TableBody';
+import TableCell from '@material-ui/core/TableCell';
+import TableHead from '@material-ui/core/TableHead';
+import TableRow from '@material-ui/core/TableRow';
+import { QueryCoordinator } from "@dolittle/queries";
+import { CaseReportsPerRegionLast7DaysQuery } from "../../Features/CaseReports/CaseReportsPerRegionLast7DaysQuery";
```
cathinkaw: This needs to be updated when the query is renamed

Source/Analytics/Read/CaseReports/CaseReportsPerRegionLast4WeeksQuery.cs
agnetedjupvik: Aha, yes, this must be updated. I think it's the command `dolittle add query CaseReportsPerRegionLast4Weeks` – let that replace this one. The artifact ID will be updated, in the case that someone is using that manually – for everything else it should be updated. This one also needs to be removed from `artifacts.json`.

Source/Analytics/Web/src/components/Reports/ReportsPerHealthRiskPerRegionLast4Weeks.js
```diff
@@ -0,0 +1,109 @@
...
+            let totalReports = region.days0to6 + region.days7to13 + region.days14to20 + region.days21to27;
+            row.push(totalReports);
+
+            rows.push(row);
+        });
+        return rows
+    }
+
+    renderRows() {
+        const currentHealthRisk = this.state.regionsForHealthRisk.find(r => r.healthRiskName == this.props.healthRisk);
```
agnetedjupvik: I'm afraid it might be difficult to understand what `this.props.healthRisk` indicates. This is the only way it is mentioned. Maybe a renaming could be smart, such as `selectedHealthRisk`, or some [PropTypes](https://reactjs.org/docs/typechecking-with-proptypes.html) validation to inform on what this value should be.

Source/Analytics/Web/src/components/Reports/ReportsPerHealthRiskPerRegionLast4Weeks.js
```diff
@@ -0,0 +1,109 @@
...
+                    <TableRow key={rowName}>
+                        <TableCell
+                            className="cell"
+                            align="left">{rowName}
+                        </TableCell>{/* Special treatment of region name column */}
+                        {row.map((numCases, i) => (
+                            <TableCell key={i}
+                                className="cell"
+                                align="right"
+                                style={numCases === 0 ? { color: "#B5B5B5" } : {}}>
```
agnetedjupvik: Use the `.nodata` class here as well?

PR #1169, opened 2019-07-30
Fixed headlines in country overview

gardnk:
Headlines now match the sketch for country overview. I took the liberty of simplifying the select health risk from the sketch, I think it looks better like this but am open to changing it.

Also gathered all the styling for country overview and lightweight overview into main.scss.

Should be merged after #1167


Source/Analytics/Web/src/assets/main.scss
```diff
@@ -1,3 +1,98 @@
...
+    font-size: 16px;
+    font-family: "Lato", "Helvetica Neue", sans-serif;
+}
+
 body {
-  margin: 0;
+    margin: 0;
+}
+
+$rc-red: #ca0000;
```
cathinkaw: these should be defined in a colors.scss that we can include in the different scss-files that use colors.

Source/Analytics/Web/src/assets/main.scss
```diff
@@ -1,3 +1,98 @@
...
+$light-grey: #f0f0f0;
+$medium-grey: #cfcdcc;
+
+$screen-xs: 480px;
```
cathinkaw: these aren't needed?

Source/Analytics/Web/src/assets/main.scss
```diff
@@ -1,3 +1,98 @@
...
+    &.header {
+        font-size: 28px;
+    }
+
+    &.description {
+        font-size: 18px;
+    }
+}
+
+.overview-top {
```
cathinkaw: This shouldn't be in main since it is specific for countryoverview? I would rather have it in a country-overview.scss file?

Source/Analytics/Web/src/assets/main.scss
```diff
@@ -1,3 +1,98 @@
...
+}
+
+.lightweightContainer {
+    margin: 1rem;
+}
+
+.lightweight {
+    text-align: center;
+
+    &.header {
```
cathinkaw: Are these classnames dependent on uimaterial? if not I think the names should be lightweight-header and lightweight-description

cathinkaw: And if I am correct the &. will mean that you are hitting something with both classnames ie class="lightweight header" and not <div class="lightweight"><div class="header">.... (as it looks like you want to hit when looking at the html)

Source/Analytics/Web/src/components/Lightweight/LightweightAreaOverview.js
```diff
@@ -11,7 +10,7 @@ export default class LightweightAreaOverview extends Component {
 
     render() {
         return (
-            <div>
+            <div className="lightweightContainer">
                 <CBSNavigation activeMenuItem="analytics/" />
                 <div className="lightweight">
                     <h2 className="header">Light Area Overview</h2>
```
cathinkaw: change according to comments made in scss-file

Source/Analytics/Web/src/components/healthRisk/CaseReportByHealthRiskTable.js
```diff
@@ -44,46 +46,60 @@ class CaseReportByHealthRiskTable extends Component {
...
+                  return (
                     <TableRow key={this.state.healthRisks[key].healthRiskName}>
-                      <TableCell component="th" scope="row">
+                      <TableCell className="cell">
                         {this.state.healthRisks[key].healthRiskName}
                       </TableCell>
-                      <TableCell align="right" style={this.state.healthRisks[key].days0to6 === 0 ? {color: "#B5B5B5"} : {}}>
+                      <TableCell 
+                        className="cell"
+                        style={this.state.healthRisks[key].days0to6 === 0 ? {color: "#B5B5B5"} : {}}>
```
cathinkaw: should this rather be set as a conditional class? ie we can have the classes cell and cell--empty. Where cell--empty overrides the color. Same for all TableCell in the rest of the files

Source/Analytics/Web/src/assets/main.scss
agnetedjupvik: I don't think all of these variables are needed, but they are used to design for different screen sizes. I think they can be written in a smarter way though, this is very verbose.

Source/Analytics/Web/src/components/healthRisk/HealthRiskSelector.js
```diff
@@ -48,19 +47,22 @@ export default class HealthRiskSelector extends Component {
...
                         {this.renderOptions()}
                     </Select>
                 </h2>
-
-                <h5>Reports the last 7 days</h5>
-                <ReportsPerHealthRiskPerDay healthRisk={this.state.selected} />
-            </div>
+                <div className="tableContainer">
+                    <h5>Reports for the last 7 days</h5>
+                    <ReportsPerHealthRiskPerDay healthRisk={this.state.selected} />
```
agnetedjupvik: Should this be `selectedHealthRisk`? It would make a lot of sense to make this state, and the `ReportsPerHealthRiskPerDay` props match in naming.

PR #1170, opened 2019-07-30
Make Analytics menu work on all screen sizes

cathinkaw:


Source/Analytics/Web/src/components/Navigation/cbs-navigation.scss
```diff
@@ -0,0 +1,257 @@
+$link-color: rgb(206, 23, 30);
```
cathinkaw: quite a lot here is restructuring

PR #1174, opened 2019-07-31
Update grid layout

agnetedjupvik:
The country overview page has been updated with a responsive grid layout. The headlines now match the sketches, and are styled uniformly by using the native tags `<h1>`, `<h2>` and `<h3>`. 

frontpage

As for mobile scalability, this is a big improvement, but the larger tables still overflow. They can be viewed by sidescrolling.


Source/Analytics/Web/src/assets/main.scss
```diff
@@ -3,3 +3,28 @@
 html {
     font-size: 16px;
 }
+
+body {
+    margin: 0;
+}
+
+h1 {
```
cathinkaw: I think we should still use classes and not set styling directly on the html tags. It will quickly and quite easily become a need for things to be a bit different.

cathinkaw:
The comments about some of the h2 tags needing the class headerText should of course be that they need the css class name that you define :P

Source/Analytics/Web/src/components/Lightweight/LightweightAreaOverview.js
```diff
@@ -11,14 +12,19 @@ export default class LightweightAreaOverview extends Component {
...
-                <CBSNavigation activeMenuItem="analytics/#" />
-                <div className="lightweight">
-                    <h2 className="header">Light Area Overview</h2>
-                    <p className="description">This is the light version of the country overview page. If you want the normal version click <Link to="/analytics">here</Link></p>
-                </div>
-                <HealthRiskSelector />
-            </div>
+            <Grid container>
+                <Grid item xs={12}>
+                    <CBSNavigation activeMenuItem="analytics/#" />
```
gardnk: Should the menu be inside the Grid?

Source/Analytics/Web/src/components/Lightweight/LightweightAreaOverview.js
```diff
@@ -11,14 +12,19 @@ export default class LightweightAreaOverview extends Component {
...
+                </Grid>
+                
+                <Grid container item xs={12} justify="center">
+                    <div className="lightweight">
+                        <h1>Light Area Overview</h1>
+                        <p>This is the light version of the country overview page. If you want the normal version click <Link to="/analytics">here</Link></p>
+                    </div>
+                    <HealthRiskSelector />
```
gardnk: Works well on small screens but the header and HealthRiskSelector are side by side on big screens, they should be centered and one below the other as for mobile.

Source/Analytics/Web/src/components/Lightweight/LightweightAreaOverview.js
agnetedjupvik: It doesn't have to, but it can (and I think having everything within the Grid layout is nice for easy resizing). By using `xs={12}` it spans the entire width of the container (which here equals the width of the screen).
The header probably won't be resized much, and it will always span the entire container, so it doesn't make that much of a difference in this case.

Source/Analytics/Web/src/assets/main.scss
agnetedjupvik: I agree – I took out the area-specific styling and added them to classes.

Source/Analytics/Web/src/components/Lightweight/LightweightAreaOverview.js
agnetedjupvik: Well spotted:) Fixed it.

PR #1177, opened 2019-08-01
Design fixes

agnetedjupvik:
Many small fixes to the design.

The largest differences:
- Select now uses the native select module instead of Material Design's.
- Change in how dates are displayed in the "Last 7 days" table.
- Table data is centered, and uses less overriding and more of TableCell props.
- Menu is placed "inside" the grid layout on large screens.
- Removed unused components.

frontpage-01-08


Source/Analytics/Web/src/assets/main.scss
```diff
@@ -25,26 +26,19 @@ body {
...
 
 .headerText {
+  font-family: "Lato", sans-serif;
   color: black !important; //these are marked as !important to override Google's MuiTable styling
-  font-weight: 200;
-
-  &.center {
-    text-align: center !important;
-  }
+  padding: 0.3rem !important;
```
cathinkaw: Do we need to have this? 😢

Source/Analytics/Web/src/assets/main.scss
```diff
@@ -25,26 +26,19 @@ body {
...
 .cell {
-  &.center {
-    text-align: center;
-  }
-}
-
-.cell--empty {
-  color: #b5b5b5 !important;
+  font-family: "Lato", sans-serif;
+  padding: 0 1rem !important;
```
cathinkaw: Same here

Source/Analytics/Web/src/components/Reports/ReportsPerHealthRiskPerDay.js
```diff
@@ -84,25 +84,42 @@ export default class ReportsPerHealthRiskPerDay extends Component {
...
             </TableRow>)
         );
     }
 
     renderDays() {
         return this.state.reportsPerHealthRisk.map(report => {
-            const day = new Date(report.timestamp).toDateString();
-            return <TableCell className="headerText" key={report.id}>{day}</TableCell>
+            const day = new Date(report.timestamp);
+            return <TableCell className="headerText date" align="center" key={report.id}>
```
cathinkaw: Only show the date of this format "28.07" (not today and not the word reports)

Source/Analytics/Web/src/assets/main.scss
agnetedjupvik: Unfortunately yes 😭 It's the easiest way to override MaterialUI's own styles.
However, there are much better ways, but they require more setup. See [here](https://material-ui.com/styles/basics/).
We could add an issue for implementing one of those solutions, it could be nice to have a more uniform way of styling throughout the system.

Source/Analytics/Web/src/components/Reports/ReportsPerHealthRiskPerDay.js
agnetedjupvik: I spoke to Hannah and we agreed on this format. But I can update it to this instead.

PR #1180, opened 2019-08-02
Implemented training mode in reporting and analytics

gardnk:
When data collectors are in training mode, the reports are not stored in analytics and not stored in the same place as the current data in reporting.

I added a button in frontend to toggle between training data and live data. There are now two identical components, one for training and one for live


gardnk:
I created an event for training reports and removed all handling of data collector in training from analytics.

LastActive is only updated for data collectors not in training, because it will look strange if analytics shows active data collectors but no reports.

PR #1188, opened 2019-08-30
Navigation component cleanup

gardnk:
* Cleaned up html in navigation component
* Updated imports
* Added build scripts to automate builds


Source/Navigation/Web.React/lib/index.js
```diff
@@ -94,6 +94,9 @@ class CBSNavigation extends Component {
       // TODO: mark the dropdown elements as well as the main one
     }
 
+    if (active == '') {// admin case
```
cathinkaw: Should this also set admin as active as in the other index.js-file or is this correct?

PR #1195, opened 2019-09-04
Data collector last active in Reporting

gardnk:
The list of data collectors now shows the last active date in the format: YYYY-MM-DD


gardnk:
I updated this PR to use the correct data collector read model.

As for the heart beats, they are currently handled by adding 99 as a health risk and will trigger the same events as a regular report. They should probably be handled differently, but for now this would mean the last active would include the heart beats as well.

PR #1209, opened 2019-09-13
refresh data collector list after deletion

gardnk:
Some cleanup of delete component. The data collector list now refreshes upon deletion.


gardnk:
> Things look coherent. But how does the command in delete.ts contain the correct datacollectorid?

The data collector object is an input to the delete.ts component, so it has all the properties for the given data collector.

cathinkaw:
👍

PR #1210, opened 2019-09-13
Edit datacollector map

gardnk:
Render map after view init and updated scss to make map render correctly

### DIFF
--- a/Source/Reporting/Web/src/app/Management/DataCollectors/edit/edit.ts
+++ b/Source/Reporting/Web/src/app/Management/DataCollectors/edit/edit.ts
@@ -29,8 +29,8 @@ export class Edit implements OnInit, AfterViewInit {
     error = false;
     dataCollector: DataCollector;
     phoneNumberString = '';
-    lat = 9;
-    lng = 44;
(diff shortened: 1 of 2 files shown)


Source/Reporting/Web/src/app/Management/DataCollectors/edit/edit.ts
```diff
@@ -22,13 +22,15 @@ import * as L from 'leaflet';
...
 })
-export class Edit implements OnInit {
+export class Edit implements OnInit, AfterViewInit {
     queryCoordinator: QueryCoordinator;
     commandCoordinator: CommandCoordinator;
 
     error = false;
     dataCollector: DataCollector;
     phoneNumberString = '';
+    lat = 9;
```
SikFar: Preferebly you want to declare variables private or public, so that we can mange complexity. Ideally it is something that should have more focus forward. Its just a question of code quality :)

gardnk: Great that you got my back :)

PR #1216, opened 2019-09-27
Analytics aggregation fix

### LINKED ISSUE #1214: Analytics data not aggregated
In [CaseReportEventProcessor](https://github.com/IFRCGo/cbs/blob/master/Source/Analytics/Read/CaseReports/CaseReportEventProcessor.cs) in Analytics, the aggregation that is done if the health risk is already reported on for that day/week is not stored. This results in the data in the tables not being correct. 

The bug is in these two functions: 
* InsertPerHealthRiskAndRegionForDay
* InsertPerHealthRiskAndRegionForComing4Weeks

gardnk:
I updated the data model for CaseReportsPerRegionLast4Weeks to make it easier to update and aggregate data correctly.

Currently both of the tables in analytics are not aggregated because the event processor is not updating an existing entry in the database. That means that each day/week in the table is consisting of a single report only.

I have made scripts to clean the data already in the database to relaunch analytics from scratch and make it work with correct data.

Fixes #1214

### DIFF
--- a/Source/Analytics/Read/CaseReports/CaseReportEventProcessor.cs
+++ b/Source/Analytics/Read/CaseReports/CaseReportEventProcessor.cs
@@ -160,8 +160,7 @@ namespace Read.CaseReports
                 {
                     if (reportsPerRegion.TryGetValue(region.Name, out int totalReports))
                     {
-                        totalReports += numberOfReports;
-                        reportsPerRegion[region.Name] = totalReports;
+                        reportsPerRegion[region.Name] = totalReports + numberOfReports;
                     }
                     else
                     {
--- a/Source/Analytics/Read/CaseReports/CaseReportEventProcessor.cs
+++ b/Source/Analytics/Read/CaseReports/CaseReportEventProcessor.cs
@@ -159,11 +159,13 @@ namespace Read.CaseReports
                     if (reportsPerRegion.TryGetValue(region.Name, out int totalReports))
                     {
                         totalReports += numberOfReports;
+                        reportsPerRegion[region.Name] = totalReports;
                     }
                     else
                     {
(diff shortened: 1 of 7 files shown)


Source/Analytics/Read/CaseReports/CaseReportEventProcessor.cs
```diff
@@ -159,11 +161,13 @@ public void InsertPerHealthRiskAndRegionForDay(CaseReport report, HealthRisk hea
                     if (reportsPerRegion.TryGetValue(region.Name, out int totalReports))
                     {
                         totalReports += numberOfReports;
+                        reportsPerRegion[region.Name] = totalReports;
```
cathinkaw: can this be simplified to be one line? I can't see that totalReports is being used later.
reportsPerRegion[region.Name] = totalReports + numberOfReports?

Source/Analytics/Web/src/components/Reports/ReportsPerHealthRiskPerRegionLast4Weeks.js
```diff
@@ -50,16 +50,18 @@ class ReportsPerHealthRiskPerRegionLast4Weeks extends Component {
...
+        const currentHealthRisk = this.state.regionsForHealthRisk[this.props.selectedHealthRisk];
 
         if (currentHealthRisk != null) {
-            return currentHealthRisk.regions.map(region => {
-                const totalReports = Object.values(region).reduce(function (acc, val) {
-                    if (!isNaN(val))
-                        return acc + val;
-                    return acc
-                }, 0);
+            const regions = Object.keys(currentHealthRisk.regions);
```
cathinkaw: indentation

